Issue #1, opened 2016-03-25
Windows builds are failing


ocefpaf:
HDF5 is failing on Windows.

```
BUILD END: hdf5-1.8.16-vc9_0 
TEST START: hdf5-1.8.16-vc9_0 
Fetching package metadata: ........
Solving package specifications: ....
Error: The following specifications were found to be in conflict:
  - hdf5 1.8.16 vc9_0
Use "conda info <package>" to see the dependencies for each package.
Command exited with code 1
```

jakirkham:
That is a very special error. I'm not really sure how we got into this situation.

jakirkham:
I dropped it in the "I need help on Windows" issue so hopefully someone with a dev environment who knows more can figure this out. Were there any recent releases of `conda` as this was working before?

jakirkham:
Oh, maybe we need `python` as a testing requirement. Thoughts? I'll draw up a PR. Feedback welcome. :smile:

ocefpaf:
> Oh, maybe we need `python` as a testing requirement. Thoughts? I'll draw up a PR. Feedback welcome. :smile: 

Makes sense... I guess...

If that works can you update the wiki?

https://github.com/conda-forge/staged-recipes/wiki/VC-features

jakirkham:
Still weird that it worked before though. Maybe there is some subtle difference between CI here and in staged-recipes?

msarahan:
Python should not be required in tests if it is required in run requirements.  However, there is one crazy issue:  Test requirements are NOT pinned the same way that run requirements are: https://github.com/conda/conda-build/issues/835

I can't explain the inconsistent behavior, but it might be related to that issue.

jakirkham:
Slowly coming back to me. We ran into [this issue with GLPK]( https://github.com/conda-forge/staged-recipes/pull/141#issuecomment-198520791 ). Maybe I need to tweak the recipe here to match. That or will need @pelson's CI improvements. Might want to wait to see how GLPK gets fixed before implementing something here. @jjhelmus @pelson, do either of you have any thoughts on this point?

jjhelmus:
At least for GLPK adding `python` as a [run requirement](https://github.com/conda-forge/glpk-feedstock/blob/master/recipe/meta.yaml#L25) on Winodws fixed the build issue.  I agree with the desire for a more elegant solution either with a msvc package or in conda build itself but until then the python run requirement will do the trick.

ocefpaf:
[...]

PR #2, opened 2016-03-25
Try to fix Windows build

jakirkham:
Fixes https://github.com/conda-forge/hdf5-feedstock/issues/1

Tries to fix the Windows build. In particular, the testing section seems not to be working because HDF5 can't be installed. Here we try to fix it by making sure the right feature is enabled (`vc*`) by installing Python. Hopefully, it ends up being the correct Python version.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -44,6 +44,7 @@ test:
     {% set hdf5_unix_cmds = [
         "h5c++",
         "h5cc",
+        "h5perf_serial",
         "h5redeploy"
     ] %}
     {% for each_hdf5_unix_cmd in hdf5_unix_cmds %}
@@ -62,7 +63,6 @@ test:
         "h5jam",
         "h5ls",
         "h5mkgrp",
-        "h5perf_serial",
         "h5repack",
         "h5repart",
         "h5stat",
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,7 +20,7 @@ source:
     - test_Makefile.in.patch
 
 build:
-  number: 1
+  number: 2
   features:
     - vc9     # [win and py27]
     - vc10    # [win and py34]
@@ -40,12 +40,20 @@ test:
     - python    # [win]
 
   commands:
+    # Verify UNIX CLI tools.
+    {% set hdf5_unix_cmds = [
+        "h5c++",
+        "h5cc",
+        "h5redeploy"
+    ] %}
+    {% for each_hdf5_unix_cmd in hdf5_unix_cmds %}
+    - command -v {{ each_hdf5_unix_cmd }}   # [unix]
+    {% endfor %}
+
     # Verify CLI tools.
     {% set hdf5_cmds = [
         "gif2h5",
         "h52gif",
-        "h5c++",
-        "h5cc",
         "h5copy",
         "h5debug",
         "h5diff",
@@ -55,7 +63,6 @@ test:
         "h5ls",
         "h5mkgrp",
         "h5perf_serial",
-        "h5redeploy",
         "h5repack",
         "h5repart",
         "h5stat",
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -36,6 +36,9 @@ requirements:
     - zlib
 
 test:
+  requires:
+    - python    # [win]
+
   commands:
     # Verify CLI tools.
     {% set hdf5_cmds = [


pelson:
Making python a test requirement even though it isn't a test requirement is a case of the tail wagging the dog, and I'd like to avoid doing it if we can.

jakirkham:
Honestly, with how VC features are setup, I don't really see any way to avoid it. If we don't make Python a test dependency, we don't have any way of ensuring the right VC is selected for testing. This results in an inability to install the package at all and non-sense errors like this one ( https://github.com/conda-forge/rtree-feedstock/issues/1 ) that @ocefpaf described.

The only other way for us to solve this problem here is to make Python a build and runtime requirement, which feels a bit silly as HDF5 has no actual dependency on Python. This is what @jjhelmus did for GLPK. It looked like you guys were moving towards making Python a testing dependency and dropping the runtime dependency, which is where I drew my inspiration for this from ( ttps://github.com/conda-forge/glpk-feedstock/pull/1 ).

So, if we don't like either strategy, I think we are out of options.

Really, the problem here is we should be able to select VC directly and not indirectly by installing Python. However, this is a larger design problem that we should really put our minds together on and then propose or selecting an existing proposal to conda-build that will correct this.

In the interim, we really need to pick which of these two solutions we like and stick with it so that we can at least have a Windows builds that people can use now.

pelson:
> Honestly, with how VC features are setup, I don't really see any way to avoid it. 

Agreed. I think it is time to consider a VC package which tracks the VC numbers (I have no idea why this wasn't done for Python in the first place, any ideas @msarahan).

In the meantime, go for the python option and we will have to unpick the whole thing once we have a solution. :+1:

jakirkham:
Sorry, just to clarify, which Python option would you prefer I use for now?

1. Build and Runtime dependency.
2. Build and Testing dependency.

pelson:
If 2 works, 2. We don't want to needlessly install Python if we don't have to.

jakirkham:
Alright, it does seem to work, but my tests are broken on Windows. However, that is an orthogonal problem.

pelson:
> However, that is an orthogonal problem.

This would seem like a good place to fix them though, otherwise we wont get new binaries anyway...

jakirkham:
That's the plan. :wink: Had to first get a build that would run the tests. :smile:

msarahan:
I think the main reason why VC stuff was rolled into Python version was the assumption that people would always follow the community "standard" association of VC version with Python version.  It really should have been split out into its own metapackage, but then I'm not completely clear on how it gets activated, or what the implications are for any/all recipes needing to be modified to line up the feature.  I'd like to move towards @ukoethe's proposal from https://github.com/conda/conda/issues/1959 (which he has implemented in ilastik) but the resistance to Jinja internally is a hurdle.

pelson:
> I'd like to move towards @ukoethe's proposal from conda/conda#1959 (which he has implemented in ilastik) but the resistance to Jinja internally is a hurdle.

I'm not as convinced by the proposal - mainly because I don't believe features to be the best solution (we have no need to install different VC built executables into the same environment AFAIK, and *version* tracking is in many ways superior to *named* features).

I'd be quite happy if I could specify msvc as a build dependency (which is itself just a metapackage which installs nothing), and that build dependency were tracked into the built distribution (as is done with Python, numpy, R, perl and lua currently). That is why I followed up with https://github.com/conda/conda-build/pull/747 as it is my belief that will be the healthiest way of tracking dependencies (not just msvc, others too). If there is anything you can do to help that review along (even if that means outright rejection) I'd be really grateful - I'm currently fighting the urge to fork conda-build for this one reason.

I'm conscious that in the past there have been a lot of comments on the subject, and I don't think it would be fair to continue that in this issue. I'm happy enough to have a discussion on the matter - if necessary, we could always do that on the conda or conda-forge mailinglists.

ukoethe:
@msarahan I'm pleased that you like PR https://github.com/conda/conda/issues/1959, but I'd like to point out that one can do a lot better - be more user friendly and elegant - by enhancing conda-build itself (PR https://github.com/conda/conda/issues/1959 was intentionally designed to be non-intrusive on conda-build). I proposed PR https://github.com/conda/conda-build/pull/848 as a first step in this direction. Please take a look!

In fact, I'm pretty certain that functionality similar to the 'x.x' notation from https://github.com/conda/conda-build/pull/747 can be realized on the basis of PR https://github.com/conda/conda-build/pull/848 with much less programming (e.g. by jinja filters).

msarahan:
@pelson sorry https://github.com/conda/conda-build/pull/747 has been neglected.  Please chime in on https://github.com/conda/conda-build/pull/848, and I'll try to help push one of the proposals forward this week.

ocefpaf:
@jakirkham this is a though one for AppVeyor and, since you are waiting for https://github.com/conda-forge/conda-smithy/pull/107, do you mind if I cancel AppVeyor here for now?

jakirkham:
Actually, I spoke incorrectly. It looks like the build just hung mysteriously after cloning.

jakirkham:
So, I have done a build of this in my own AppVeyor project and it has [passed]( https://ci.appveyor.com/project/jakirkham/hdf5-feedstock/build/1.0.5 ). As can be seen, both have the same commit. As this could be merged with a fast-forward merge, there is nothing weird that would be added by `master` that would need to be tested in say a merge commit. In other words, we are confident that these changes here work. Even though AppVeyor has not finished here, we will go ahead and merge as it has been verified elsewhere and is equivalent to the change here.

PR #10, opened 2016-05-12
Pin

ocefpaf:
Pinning zlib


jakirkham:
Had failures. Though they look like issues downloading from anaconda.org. We must really be pushing that service to the limits. Restarting...

recipe/meta.yaml
```diff
@@ -31,13 +31,13 @@ requirements:
...
-    - zlib
+    - zlib 1.2*
   run:
-    - zlib
+    - zlib 1.2*
 
 test:
   requires:
-    - python    # [win]
+    - python {{ environ['PY_VER'] + '*' }}  # [win]
```
jakirkham: Glad to see the linter can handle this now. Thanks @ocefpaf. 😄

jakirkham: Though, I should add, we haven't needed to pin Python previously here as it has pulled in the right one because of the `zlib` dependendency. That being said, I'm ok with this change. 👍

PR #18, opened 2016-05-24
WIP: Add HDF5 1.8.15

jakirkham:
Adds HDF5 1.8.15 for packaging.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.8.16" %}
+{% set version = "1.8.15" %}
 
 package:
   name: hdf5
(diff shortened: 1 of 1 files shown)


jakirkham:
Weird. It is missing the binaries on Windows. Any thoughts?

msarahan:
Nope.  Weird.  Some of them are present in the 1.8.15 build, but missing in 1.8.15.1.  I can't believe the HDF5 team would be moving or renaming EXE's.  They must not be getting built somehow.  I'll take another look at the build log to see if there are any non-catastrophic warnings or errors leading to this.

msarahan:
It is there, just has a different name.  I don't know why this is happening without digging in CMake.  Sorry, can't do that right now.  But your build targets are ending up with different names depending on build parameters, somehow.  https://ci.appveyor.com/project/conda-forge/hdf5-feedstock/build/1.0.34/job/i0rfant6280c8y2p#L3153  HTH.

jakirkham:
I thought that was the case, but then I saw [this]( https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.1/src/unpacked/windows/tools/ ). They provide an executable and an executable with `dll` at the end too for every CLI tool on Windows. So, we are definitely missing stuff, but I don't know why.

jakirkham:
To summarize, there should be executables that do not end in `dll` and that has been the case since HDF5 1.8.1. So, HDF5 1.8.15 should be no different. We don't see this problem with 1.8.16 or 1.8.17.

jakirkham:
@patricksnape, do you have any idea why these executables would be missing on Windows?

jakirkham:
This is no longer needed and won't be supported. Closing...

PR #21, opened 2016-05-30
Update to 1.8.17

jakirkham:
Updates to HDF5 1.8.17 as it is now out.


jakirkham:
As long as this passes, it is good to merge IMHO.

jakirkham:
Passing. Going to merge so we have 1.8.17 available. 😄

PR #22, opened 2016-06-11
Use the toolchain

jakirkham:
Try switching over to using the toolchain to build HDF5. This should result in binaries using `libc++` on Mac.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,7 +20,7 @@ source:
     - test_Makefile.in.patch
 
 build:
-  number: 0
+  number: 1
   features:
     - vc9     # [win and py27]
     - vc10    # [win and py34]
@@ -29,7 +29,9 @@ build:
 
 requirements:
   build:
-    - python    # [win]
+    - toolchain
+    - python         # [win]
+    - libtool        # [unix]
     - cmake >=3.1
     - zlib 1.2*
   run:
@@ -88,6 +90,11 @@ test:
     - test -f $PREFIX/lib/lib{{ each_hdf5_lib }}.so                          # [linux]
     - if not exist %PREFIX%\\Library\\lib\\{{ each_hdf5_lib }}.lib exit 1    # [win]
     - if not exist %PREFIX%\\Library\\bin\\{{ each_hdf5_lib }}.dll exit 1    # [win]
+    {% endfor %}
+
+    # Inspect linkages of HDF5.
+    {% for each_hdf5_lib in hdf5_libs %}
+    - otool -L $PREFIX/lib/lib{{ each_hdf5_lib }}.dylib                      # [osx]
     {% endfor %}
 
 about:
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+# FIXME: This is a hack to make sure the environment is activated.
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+source activate "${CONDA_DEFAULT_ENV}"
+
+if [ "$(uname)" == "Darwin" ]
+then
+    export CXX="${CXX} -stdlib=libc++"
+fi
+
 ./configure --prefix="${PREFIX}" \
     --enable-linux-lfs \
     --with-zlib="${PREFIX}" \


jakirkham:
cc @gillins @ocefpaf

jakirkham:
It looks like this works. Weirdly AppVeyor didn't run though.

jakirkham:
Alright, now we got AppVeyor. Though it isn't really affected. It would be nice to see it builds ok before merging something.

jakirkham:
Please share your thoughts on this guys. I think it may solve some issues you have experienced based on @gillins recent feedback.

gillins:
This didn't make any difference when I test on my Mac:
```
$ otool -L ~/miniconda3/lib/libhdf5_cpp.dylib 
/Users/sam/miniconda3/lib/libhdf5_cpp.dylib:
	@rpath/libhdf5_cpp.12.dylib (compatibility version 13.0.0, current version 13.0.0)
	@rpath/libhdf5.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 104.1.0)
```

Are you sure you meant to include `toolchain` rather than `libgcc`? `toolchain` doesn't contain any libs.

Just just checked some of the other dependencies of `gdal` and the following packages also depend on system `libstdc++`: `xerces-c` (from defaults), `sqlite3` (just needs `libgcc_s.dylib`), `geos`, `icu` and `kealib`.

I think we need to have a clear plan before doing anything here considering how wide ranging it will likely be...

recipe/meta.yaml
```diff
@@ -42,6 +43,9 @@ test:
     - python 3.5.*  # [win and py35]
 
   commands:
+    # Inspect linkages of HDF5.
+    - conda inspect linkages -n _test hdf5  # [unix]
```
jakirkham: This should tell us how this is linked.

jakirkham:
> This didn't make any difference when I test on my Mac

I've added a command to show us what the linkages look like on Travis. If it looks wonky, we can try to explore this more. If it is clean though, my recommendation would be to go ahead with this to eliminate one potential cause of problems and make this compliant.

> Are you sure you meant to include `toolchain` rather than `libgcc`?

Yes.

> `toolchain` doesn't contain any libs.

It doesn't need to. It uses system libraries. The package merely exists to configure the system compilers to work as intended.

> Just just checked some of the other dependencies of gdal and the following packages also depend on system libstdc++

> `xerces-c` (from defaults)

There has been a PR ( https://github.com/conda-forge/staged-recipes/pull/194 ) to add this that has been lingering for some time. I have pinged it a few times, but have not heard back from @groutr. Someone should take it over to get it in. Honestly, I think you are the most qualified to do this @gillins (if you have interest).

> `sqlite3` (just needs `libgcc_s.dylib`)

Using `libgcc_s.dylib` shouldn't cause a problem, but I don't know why we need it. Could you please add an issue reminding me to take a look? Feel free to link it back here for an example.

> `geos`, `icu`, `kealib`

Yes, these should be fixed. Again please add an issue there to use the `toolchain` and link this as an example.

recipe/meta.yaml
jakirkham: Unfortunately, this fails due to this issue ( https://github.com/conda/conda-build/issues/836 ).

jakirkham: Switching to `otool` instead.

gillins:
Doing all these libs is a big job. I just want to make sure there is a consensus to do all this before launching in...

And it will take a while and OSX is broken in the meantime.

jakirkham:
I disagree that it is so challenging. Take a look at the changes proposed in this PR. The reason we add the `toolchain` package was to try and eliminate the guess work.

Also, we had agreed on using `clang` for Mac before and most packages already do. The smart move seems to fix the ones that are not compliant, which sounds like a couple and move forward.

jakirkham:
I am seeing the same linkage issue on Travis CI that you are seeing, @gillins. It can be seen with the [log]( https://s3.amazonaws.com/archive.travis-ci.org/jobs/137134840/log.txt ) associated with this [build]( https://travis-ci.org/conda-forge/hdf5-feedstock/builds/137134839 ). It is a bit strange as there doesn't seem to be any indication that the linkage will occur until it has already happened and is being fixed by `conda-build`. Am trying to give the compiler the `-lc++` argument to see if that might help.

jakirkham:
Passing `-lc++` did not seem to fix it. Trying to set similar flags for `CFLAGS`.

jakirkham:
So, that didn't work either. We then tried using our `libtool` package, which didn't help. Going to try forcing C++11 support. This should at least crash any `gcc` step that is hanging around.

jakirkham:
Still no luck. Can you see any reason for it to still be linked to `libstdc++`, @gillins? I'm debating manually removing this linkage and writing a C++ test to make sure it works ok. It seems very strange as there is no reference in the log to this until the end and no where in the source directory do I see any references to this library other than a doc about MinGW.

gillins:
Sorry @jakirkham I wish I was more use on this. I only recently bought a second hand Mac for the purposes of building and debugging Conda packages and still don't really know my way around.

I did come across this:
```
Before OS X 10.9.x:
    The default was libstdc++ (using clang++ -o hello.out hello.cpp) would have worked fine.
OS X 10.9.x:
    The default is libc++ (as you know the including flag -stdlib=libstdc++ links your project correctly).
```
(from : http://stackoverflow.com/questions/23666754/default-to-lib-stdlibc-for-clang-on-mac-os-mavericks-with-xcode) recently when I was trying to build something against Conda packages. 
I added `-stdlib=libstdc++` to my CXXFLAGS at that seemed to fix it. I think Conda tries to build for the older case (pre 10.9.x) hence linking to `libstdc++`.

@danclewley is the guy I usually ask about this stuff. Dan - do you have any more to add?

gillins:
I should just say I really don't mind how this is resolved as long as there is only one `libstdc++` etc linked in. I'll be away from my computer for a few days so I might be slow in responding to anything so don't wait for me.
Hopefully some Mac people can weigh in about what is best practice here.

danclewley:
@gillins nothing to add at the moment. I'll have a poke around on my machine and report back.

jakirkham:
Just to explain the issue I am seeing simply, @danclewley, the change here should just result in `libhdf5_cpp.dylib` being linked against `libc++.dylib`. However, it is getting linked against `libc++.dylib` and `libstdc++.dylib`. This smells funky to me. If you have any idea why and/or how to fix it, would be really interested in hearing your thoughts. Happy to answer any questions you may have.

jakirkham:
We also set these [environment variables]( https://github.com/conda-forge/toolchain-feedstock/blob/66ce00916b26aecda2d7ddb9bef381c31881208d/recipe/activate.sh#L5-L17 ) in the [`source` statement]( https://github.com/conda-forge/hdf5-feedstock/pull/22/files#diff-44a73bcc045c193c3bd45da87994b03bR9 ).

jakirkham:
Not to rush you (merely curious). Have you had a chance to look at this, @danclewley?

danclewley:
@jakirkham not yet. Hoping to get round to it soon. I don't have a good idea of why it is happening so and have a feeling it is going to take quite a lot of digging.

jakirkham:
No worries.

Yeah, I have that feeling too. 😒

gillins:
OK I've had a bit of a look into this and it seems `libtool` is actually removing `-stdlib=libc++` from the link command. It is leaving in `-lc++` and hence the library ends up being linked to both `libstdc++` and `libc++`. 

I can't work out why or how `libtool` is doing this, but I suspect this may have always been a problem with `autoconf` based packages. It is possible the `=` is confusing it.

What worked for me was to set `CXX="clang++ -stdlib=libc++"`. If this seems a sensible way to go I'll submit a PR to the toolchain-feedstock. Then we will have to rebuild all `autoconf` based builds. I can't guarantee we won't find other problems tho :grimacing: 

Thoughts?

jakirkham:
I'm not sure that `libtool` is the cause either. Using the system `libtool` (not installing our `libtool` package) results in the exact same linkage. It looks like the current builds are getting this [argument]( https://travis-ci.org/conda-forge/hdf5-feedstock/builds/137223989#L1367 ), but it is unclear what they are doing with it.

Honestly, I think there is something funky with HDF5 itself where it is forcing this behavior. That's the hunch I got from looking at the build files before, but I'm still not exactly sure how or why this behavior occurs or how to disable it. We should really try to understand this and convey the problem to upstream so that it can be fixed.

gillins:
Have you tried doing a HDF5 build using `cmake`?

`gdal` (another `libtool` package) is linking in `libstdc++`, so I was hopeful this was the problem. However looking closer I see it does **not** link in `libc++`, so it is slightly different.

gillins:
FYI, here is the output of a verbose build of the C++ bindings:
```
/bin/sh ../../libtool  --tag=CXX   --mode=link clang++  -mmacosx-version-min=10.7 -stdlib=libc++ -m64 -version-info 12:0:0 -L/Users/sam/miniconda3/envs/_build_placehold_placehold_placehold_placehold_placeh/lib  -headerpad_max_install_names -mmacosx-version-min=10.7 -stdlib=libc++ -lc++ -o libhdf5_cpp.la -rpath /Users/sam/miniconda3/envs/_build_placehold_placehold_placehold_placehold_placeh/lib H5Exception.lo H5IdComponent.lo H5Library.lo H5Attribute.lo H5Location.lo H5Object.lo H5PropList.lo H5FaccProp.lo H5FcreatProp.lo H5DcreatProp.lo H5DxferProp.lo H5OcreatProp.lo H5DataType.lo H5DataSpace.lo H5AbstractDs.lo H5AtomType.lo H5PredType.lo H5EnumType.lo H5IntType.lo H5FloatType.lo H5StrType.lo H5ArrayType.lo H5VarLenType.lo H5CompType.lo H5DataSet.lo H5CommonFG.lo H5Group.lo H5File.lo ../../src/libhdf5.la -lz -ldl -lm 

libtool: link: clang++ -dynamiclib -Wl,-undefined -Wl,dynamic_lookup -o .libs/libhdf5_cpp.12.dylib  .libs/H5Exception.o .libs/H5IdComponent.o .libs/H5Library.o .libs/H5Attribute.o .libs/H5Location.o .libs/H5Object.o .libs/H5PropList.o .libs/H5FaccProp.o .libs/H5FcreatProp.o .libs/H5DcreatProp.o .libs/H5DxferProp.o .libs/H5OcreatProp.o .libs/H5DataType.o .libs/H5DataSpace.o .libs/H5AbstractDs.o .libs/H5AtomType.o .libs/H5PredType.o .libs/H5EnumType.o .libs/H5IntType.o .libs/H5FloatType.o .libs/H5StrType.o .libs/H5ArrayType.o .libs/H5VarLenType.o .libs/H5CompType.o .libs/H5DataSet.o .libs/H5CommonFG.o .libs/H5Group.o .libs/H5File.o   -L/Users/sam/miniconda3/envs/_build_placehold_placehold_placehold_placehold_placeh/lib ../../src/.libs/libhdf5.dylib -lc++ -lz -ldl -lm  -mmacosx-version-min=10.7 -m64 -mmacosx-version-min=10.7   -install_name  /Users/sam/miniconda3/envs/_build_placehold_placehold_placehold_placehold_placeh/lib/libhdf5_cpp.12.dylib -compatibility_version 13 -current_version 13.0 -Wl,-single_module
```
The `stdlib=libc++` bit seems to be dropped when it calls `clang++`.

gillins:
I'm getting the same behaviour with `geos` (once I updated it to use `toolchain`)....

jakirkham:
> Have you tried doing a HDF5 build using `cmake`?

IIRC @groutr ran into some really subtle issues downstream trying to do this. So, I'd rather not try that.

FWIW I also hit this problem with `jsoncpp` and it uses `cmake`. Thus, I'm not thinking that will fix it.

> OK I've had a bit of a look into this and it seems `libtool` is actually removing `-stdlib=libc++` from the link command. It is leaving in `-lc++` and hence the library ends up being linked to both `libstdc++` and `libc++`.

After some time investigating this more, I think you're theory about `libtool` may be right, @gillins. I just don't think it is unique to the conda-forge `libtool` as it can be reproduced with the system `libtool` as well.

> What worked for me was to set `CXX="clang++ -stdlib=libc++"`. If this seems a sensible way to go I'll submit a PR to the toolchain-feedstock. Then we will have to rebuild all `autoconf` based builds. I can't guarantee we won't find other problems tho :grimacing:

That sort of things makes me nervous, but if it works. Great. However, I just tried this with `cmake` in a different recipe and it did not like it.

```
CMake Error at CMakeLists.txt:4 (PROJECT):
  The CMAKE_CXX_COMPILER:

    clang++ -stdlib=libc++

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

gillins:
> However, I just tried this with cmake in a different recipe and it did not like it.

OK that's no good. Seems we aren't getting anywhere fast.
I wonder whether we are doing a sensible thing... Should we be just trying to link to `libstdc++` and **not** `libc++` instead? Or increasing the minimum OSX version to 10.9?
Do you know what the reasoning behind the current settings are?

jakirkham:
> OK that's no good. Seems we aren't getting anywhere fast.

We could try adopting your proposed workaround in the few places its needed. That seems totally sensible to me instead of trying to find the perfect solution. We should just no its scope. Right now that is no `cmake`. Eventually, we do have to figure out how to absorb this into the `toolchain` at some point so we don't have to hunt down more issues like this in the future.

> Do you know what the reasoning behind the current settings are?

> I wonder whether we are doing a sensible thing... Should we be just trying to link to `libstdc++` and **not** `libc++` instead?

That's pretty much a no go. We need `libc++` for two reasons. The first is we need C++11 support. The second is it appears to be absent from 10.11 AFAICT. Both are pretty bad, but the latter is worse IMHO.

> Or increasing the minimum OSX version to 10.9?

There's a long discussion in this issue ( https://github.com/conda-forge/toolchain-feedstock/issues/7 ) about doing that. I'm not opposed, but this might make it harder for Continuum to reuse our recipes. Was really hoping a Continuum person could comment on that change, but we have heard nothing so far.

jakirkham:
As further verification of the idea that `libtool` is suspect, I tried to build it with the toolchain. Note that it stripped `-stdlib=libc++` from the compiler flags.

```

       host-triplet:   x86_64-apple-darwin13.4.0
       shell:          /bin/sh
       compiler:       clang
       compiler flags:  -mmacosx-version-min=10.7 -m64
       linker:         /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld (gnu? no)
       version:        libtool (GNU libtool) 2.4.6
       automake:       automake (GNU automake) 1.11
       autoconf:       autoconf (GNU Autoconf) 2.68
```

gillins:
Another option could be to write our own shell script and point `CXX` to that. The shell script would then run `clang++` with our options, but it would have to decide whether it is being used as a compiler or linker and adjust the arguments appropriately so not a trivial thing.

If I have time today I'll try and see if changing the minimum OSX version to 10.9 fixes these problems.

jakirkham:
Well, I can save you the trouble. FWICT it doesn't seem to.

jakirkham:
I'm also trying to play with `LIBTOOLFLAGS` to see if we can force it to give us the right behavior.

jakirkham:
> If I have time today I'll try and see if changing the minimum OSX version to 10.9 fixes these problems.

Sorry, I missed a few spots where it needed to be set. This looks promising.

jakirkham:
Yep, this will fix the build of HDF5. I'm going to go ahead propose it and see what people say.

jakirkham:
Proposing the bump to 10.9 as Mac OS minimum deployment target in this PR ( https://github.com/conda-forge/toolchain-feedstock/pull/9 ). This works for me locally with this change.

jakirkham:
Since you are the compiler whisperer (or at least on a roll catching my typos 😆), do you have any idea what is going on with the linkage here, @stuarteberg? Basically, it looks like the same issue we saw in PR ( https://github.com/conda-forge/staged-recipes/pull/826 ).

stuarteberg:
>do you have any idea what is going on with the linkage here

No, I'm as mystified as you are.  As a last resort, we could try using the `cmake`-based configuration option that HDF5 provides, instead of using `configure`.  Maybe cmake gets it right?

jakirkham:
Yeah, the best we came up with was `libtool` is doing something weird (stripping the `-stdlib=libc++` flag). Though I'd really want to understand why `libtool` misbehaves this way (if that is in fact the problem) as there are plenty of things that use `libtool`.

Maybe using `cmake` will avoid that. Though I had the impression that @groutr had issues with `cmake` previously. Something about segfaults with `h5py` (it was in an outdated diff so I quoted it below). ( [ref]( https://github.com/conda-forge/staged-recipes/pull/192#discussion_r57229907 ) )

> I tried using cmake, but the resulting library caused the python interpreter to segfault when h5py was imported under certain circumstances.  I'm not sure they've worked all the kinks out of their cmake scripts.  Their official binaries seem to still be built using this method.
See https://github.com/ContinuumIO/anaconda-issues/issues/687

jakirkham:
I'm going to use your workaround ( `CXX="clang++ -stdlib=libc++"` ), @gillins. Please merge if this looks ok to you.

gillins:
I'm not a maintainer of `hdf5` so can't merge, but it certainly looks ok to me.

I tried this workaround on `geos` yesterday, but it didn't seem to work there - the right flags got passed but the library still ended linked to both runtimes. I haven't had a chance to have a proper look yet.

I think https://github.com/conda-forge/toolchain-feedstock/pull/9 is still our best bet at being able to rebuild everything easily...

jakirkham:
FYI here is the result of running `otool -L` from the [log]( https://s3.amazonaws.com/archive.travis-ci.org/jobs/139114876/log.txt ). Only seeing `libc++` now.

```
+otool -L /Users/travis/miniconda3/envs/_test/lib/libhdf5.dylib
/Users/travis/miniconda3/envs/_test/lib/libhdf5.dylib:
	@rpath/libhdf5.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
+otool -L /Users/travis/miniconda3/envs/_test/lib/libhdf5_cpp.dylib
/Users/travis/miniconda3/envs/_test/lib/libhdf5_cpp.dylib:
	@rpath/libhdf5_cpp.12.dylib (compatibility version 13.0.0, current version 13.0.0)
	@rpath/libhdf5.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
+otool -L /Users/travis/miniconda3/envs/_test/lib/libhdf5_hl.dylib
/Users/travis/miniconda3/envs/_test/lib/libhdf5_hl.dylib:
	@rpath/libhdf5_hl.10.dylib (compatibility version 12.0.0, current version 12.0.0)
	@rpath/libhdf5.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
+otool -L /Users/travis/miniconda3/envs/_test/lib/libhdf5_hl_cpp.dylib
/Users/travis/miniconda3/envs/_test/lib/libhdf5_hl_cpp.dylib:
	@rpath/libhdf5_hl_cpp.11.dylib (compatibility version 13.0.0, current version 13.0.0)
	@rpath/libhdf5_hl.10.dylib (compatibility version 12.0.0, current version 12.0.0)
	@rpath/libhdf5_cpp.12.dylib (compatibility version 13.0.0, current version 13.0.0)
	@rpath/libhdf5.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

jakirkham:
> I'm not a maintainer of `hdf5` so can't merge...

Ah, sorry, for some reason I thought you were. Would you like to be added?

> ...but it certainly looks ok to me.

👍 That's what I wanted to know. 😄

> I tried this workaround on `geos` yesterday, but it didn't seem to work there - the right flags got passed but the library still ended linked to both runtimes. I haven't had a chance to have a proper look yet.

Hmm...I'll try to look at some point. Can't promise it will be today though. Could you please put what you have in a PR and ping me there, @gillins?

> I think https://github.com/conda-forge/toolchain-feedstock/pull/9 is still our best bet at being able to rebuild everything easily...

I agree. This trying to support the possibility of 10.7 even though we don't build with anything older than 10.9 is a bit silly. Now that it is causing us serious problems that are tricky to debug, it is really worth questioning whether this is worth the effort.

Given the nature of the change though, I want to really make sure people are ok with it before going ahead and I don't want to hold-up a working solution here over it. So, am going to go ahead and merge this now. We can remove the workaround later if/when we switch to 10.9.

jakirkham:
The OS X and Linux builds are out. Those are the primary ones affected by this change. Windows is pending, but is not really changes by this. Please let me know if you have more issues.

gillins:
> Would you like to be added?

Don't mind, maybe the next time you are making a change but it's not a big deal.

> The OS X and Linux builds are out.

And the result looks good to me :smile: 

> Could you please put what you have in a PR and ping me

See https://github.com/conda-forge/geos-feedstock/pull/16

I think I must have been looking at the wrong file, or had something wrong in `build.sh`. When I ran a build this morning while preparing the PR `geos` was fine (only `libc++`). So I think we are making progress :tada:

jakirkham:
I'm going to go ahead and get the ball rolling on adding you now, @gillins. See this PR ( https://github.com/conda-forge/hdf5-feedstock/pull/24 ). Thanks for helping me sort this out.

PR #24, opened 2016-06-22
Add @gillins

jakirkham:
This add @gillins as a maintainer on the `hdf5` feedstock.

I figured I would go ahead and add you now as you have been so helpful. Given your interest in getting this working, it feels silly to restrict your ability to make substantive change when we need it. Hope that is ok.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -107,5 +107,6 @@ about:
 extra:
   recipe-maintainers:
     - jakirkham
+    - gillins
     - groutr
     - ocefpaf


gillins:
Thanks @jakirkham :+1:

jakirkham:
Of course. 😄

jakirkham:
Normally our cron job script would have run by now and added you, but it appears there is some sort of backlog. Will try restarting it. Though this will exhaust or GitHub rate limit whenever it decides to run.

gillins:
It has just come through now - thanks!

jakirkham:
Awesome. Glad it went through. Turn out the Linux container builds and OS X build for the org share a queue. So, when OS X builds got backed up, the Linux container builds did too (that's where the cron job is run).

Issue #23, opened 2016-06-22
install failing due to links


david-murr:
Install is failing at
IOError: [Errno 2] No such file or directory: '/Users/xxx/anaconda/pkgs/hdf5-1.8.17-1/bin/activate'

A quick look shows that these have been installed as links (to nonexistent directories)
lrwxr-xr-x  1 xxxx  staff      37 Jun 21 19:03 activate -> /Users/travis/miniconda3/bin/activate
lrwxr-xr-x  1 xxxx  staff      34 Jun 21 19:03 conda -> /Users/travis/miniconda3/bin/conda
lrwxr-xr-x  1 xxxx  staff      39 Jun 21 19:03 deactivate -> /Users/travis/miniconda3/bin/deactivate

condo info
Using Anaconda Cloud api site https://api.anaconda.org
Current conda install:

             platform : osx-64
        conda version : 4.1.2
  conda-build version : 0+unknown
       python version : 2.7.11.final.0
     requests version : 2.9.1
     root environment : /Users/xxxx/anaconda  (writable)
  default environment : /Users/xxxx/anaconda
     envs directories : /Users/xxxx/anaconda/envs
        package cache : /Users/xxxx/anaconda/pkgs
         channel URLs : https://conda.anaconda.org/conda-forge/osx-64/
                        https://conda.anaconda.org/conda-forge/noarch/
                        https://repo.continuum.io/pkgs/free/osx-64/
                        https://repo.continuum.io/pkgs/free/noarch/
                        https://repo.continuum.io/pkgs/pro/osx-64/
                        https://repo.continuum.io/pkgs/pro/noarch/
          config file : /Users/xxxx/.condarc
    is foreign system : False

gillins:
This is quite serious as `conda` gets overwritten breaking the user's environment. I can confirm same thing on my Mac.

Is this a problem with the `toolchain` package @jakirkham ? There it seems to some copying of scripts in its `build.sh`.

jakirkham:
Hmm... I don't think so. The toolchain has UNIX scripts are named `toolchain_activate.sh` not `activate`. Also, `conda-build` normally avoids files that came from a dependency unless they are explicitly exempted (e.g. [`libgcc`]( https://github.com/conda/conda-recipes/blob/d34a9a134844f36fff885ce3cf92287efb17222c/libgcc/meta.yaml#L12 )). As we didn't add these files or exempt them, I don't think it was an error in our packaging.

[...]

PR #25, opened 2016-06-22
Unpin `conda-build`

jakirkham:
Fixes https://github.com/conda-forge/hdf5-feedstock/issues/23

This PR unpins `conda-build` on Travis CI. The mismatch between functionality regarding environments and their activation in `conda` and `conda-env` vs `conda-build` is causing some rather strange problems. In particular, it is result in symlink files getting packaged with the binaries in some cases. Switching to a newer version of `conda-build` fixes this problem as it properly ignores these activation files. Thus we unpin `conda-build` and bump the build number so we can get a new package out without the activation scripts included.

cc @gillins @david-murr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
 
       conda config --set show_channel_urls true
       conda update --yes conda
-      conda install --yes conda-build=1.20.0 jinja2 anaconda-client
+      conda install --yes conda-build jinja2 anaconda-client
       conda config --add channels conda-forge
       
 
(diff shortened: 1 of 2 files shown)


jakirkham:
FYI Travis CI on Mac is pretty [backed up]( https://www.traviscistatus.com/incidents/nf3tg08pqg2d ). May be a while before we see this build.

jakirkham:
This should be mergeable. What do you think about taking it in, @gillins?

PR #43, opened 2016-09-27
Don't use DYLD_FALLBACK_LIBRARY_PATH

astrofrog:
... because it doesn't work with recent versions of MacOS X due to the System Integrity Protection (see https://github.com/conda-forge/conda-forge.github.io/issues/238)

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -11,7 +11,7 @@ source activate "${CONDA_DEFAULT_ENV}"
 if [ "$(uname)" == "Darwin" ]
 then
     export CXX="${CXX} -stdlib=libc++"
-    export DYLD_FALLBACK_LIBRARY_PATH=$PREFIX/lib
+    export LDFLAGS="-Wl,-rpath,$PREFIX/lib"
 fi
 
 export LIBRARY_PATH="${PREFIX}/lib"
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -13,16 +13,7 @@ h5c++ -show
 h5c++ h5tutr_cmprss.cpp -o h5tutr_cmprss
 ./h5tutr_cmprss
 
-# Test Fortran compiler. For this we need to set DYLD_FALLBACK_LIBRARY_PATH to
-# make sure libgfortran gets picked up. Ideally this shouldn't be needed, but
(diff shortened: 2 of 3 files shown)


astrofrog:
@ jakirkham @gillins @groutr @ocefpaf - any objections to this?

recipe/run_test.sh
```diff
@@ -13,16 +13,7 @@ h5c++ -show
...
 
-# Test Fortran compiler. For this we need to set DYLD_FALLBACK_LIBRARY_PATH to
-# make sure libgfortran gets picked up. Ideally this shouldn't be needed, but
-# this is how the gfortran compiler works in conda - see:
-#
-#   https://github.com/ContinuumIO/anaconda-issues/issues/739
-#
-# for more details.
-
-export DYLD_FALLBACK_LIBRARY_PATH=${CONDA_PREFIX}/lib
```
jakirkham: I'm amazed we were able to drop this, but am ok with it if it works.

recipe/build.sh
```diff
@@ -11,7 +11,7 @@ source activate "${CONDA_DEFAULT_ENV}"
 if [ "$(uname)" == "Darwin" ]
 then
     export CXX="${CXX} -stdlib=libc++"
-    export DYLD_FALLBACK_LIBRARY_PATH=$PREFIX/lib
+    export LDFLAGS="-Wl,-rpath,$PREFIX/lib"
```
jakirkham: Maybe we should do a similar thing for Linux so as to avoid setting `LIBRARY_PATH`.

jakirkham:
Yep, I'm fine with this as is.

recipe/run_test.sh
astrofrog: Me too, but it all seems to work fine!

astrofrog:
@jakirkham - thanks for checking over this - just to be sure, can I go ahead and merge?

jakirkham:
Merged.

recipe/build.sh
jakirkham: Sorry must have missed this. We should have `${LDFLAGS}` at the beginning of this assignment so as to not overwrite `$LDFLAGS`. Could you please submit a follow-up PR with a build number bump @astrofrog ?

PR #44, opened 2016-09-27
WIP: Run tests on Windows

jakirkham:
Had this floating around locally. Can't remember if there was a reason not to try this. So have added it as a PR. Best case we start running the test suite on Windows. Worst case we don't. 😄


jakirkham:
Replaced by PR ( https://github.com/conda-forge/hdf5-feedstock/pull/81 ).

PR #45, opened 2016-09-30
Enable Fortran 2003

scopatz:
This is needed for some downstream packages for HDF5 < 1.10. https://github.com/mit-crpg/openmc/issues/727#issuecomment-250870760

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -25,7 +25,7 @@ export LIBRARY_PATH="${PREFIX}/lib"
             --enable-fortran2003 \
             --with-default-plugindir="${PREFIX}/lib/hdf5/plugin"
 
-make
+make -j "${CPU_COUNT}"
 make check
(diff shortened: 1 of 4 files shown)


scopatz:
Ok @astrofrog, I have added a test and it seems to work, but the appveyor queue is taking a long time. Also note the f2003 test has an f90 extension.  This was the original extension that the HDF5 group provides.

ocefpaf:
> but the appveyor queue is taking a long time. 

There is no need to wait on AppVeyor here b/c there BTW. If @astrofrog is OK with this change feel free to merge. I am :+1: to this.

scopatz:
Looks like Windows now passes too.

scopatz:
Can thos please be looked at again. It is blocking development of other recipes...

scopatz:
Thanks!

Issue #41, opened 2016-09-14
Rebuild with conda-build 2.0


dopplershift:
Ok, it looks like the 💩 has finally hit the fan--I just got this on a CI build for MetPy on AppVeyor (which is using conda-forge):

```
PaddingError: Placeholder of length '20' too short in package conda-forge::hdf5-1.8.17-vc9_3.
The package must be rebuilt with conda-build > 2.0
```

So, how do we need to proceed?

jakirkham:
Deja vu? https://github.com/conda-forge/hdf5-feedstock/issues/27

dopplershift:
If you mean old conda, here's the versions of things:
```
conda update -q conda
Fetching package metadata: ....
.Solving package specifications: .........
Package plan for installation in environment C:\Miniconda:
The following packages will be downloaded:
    package                    |            build
    ---------------------------|-----------------
    conda-env-2.6.0            |                0          488 B
    python-2.7.12              |                0        22.4 MB
    enum34-1.1.6               |           py27_0          53 KB
    ruamel_yaml-0.11.14        |           py27_0         193 KB
    conda-4.2.6                |           py27_0         411 KB
    ------------------------------------------------------------
                                           Total:        23.0 MB
The following NEW packages will be INSTALLED:
    enum34:      1.1.6-py27_0  
    ruamel_yaml: 0.11.14-py27_0
The following packages will be UPDATED:
    conda:       4.0.5-py27_0 --> 4.2.6-py27_0  
    conda-env:   2.4.5-py27_0 --> 2.6.0-0       
    python:      2.7.11-4     --> 2.7.12-0   
```

darribas:
I'm having the same issue on Appveyor for Python2.7 (can't go to Py3):

https://ci.appveyor.com/project/darribas/gds16/build/job/oy3vk0me77sfxcdu

Any help would be most welcome.

richardotis:
Same problem on py35: https://ci.appveyor.com/project/richardotis/pycalphad/build/job/bv8iho7yl4a9xf0w

tritemio:
Same problem here. AppVeyor CI for FRETBursts fails with python 2.7, 3.4 and 3.5 due to the hdf5 package from conda-forge:

https://ci.appveyor.com/project/tritemio/fretbursts/build/1.0.53

Is there a way to use default anaconda packages by default and fall-back to conda-forge only if a package cannot be found on anaconda default channel?

ocefpaf:
[...]

PR #56, opened 2017-01-14
Use conda build 2 win

ocefpaf:
Windows was never built with `conda-build 2`.

Fixes https://github.com/conda-forge/hdf5-feedstock/issues/41

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,11 +20,12 @@ source:
     - test_Makefile.in.patch
 
 build:
-  number: 8
+  number: 9
+  skip: True  # [py==36]
   features:
     - vc9     # [win and py27]
     - vc10    # [win and py34]
-    - vc14    # [win and py35]
+    - vc14    # [win and py>=35]
   detect_binary_files_with_prefix: True
 
 requirements:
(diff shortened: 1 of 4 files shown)


recipe/meta.yaml
```diff
@@ -20,11 +20,12 @@ source:
...
 
 build:
-  number: 8
+  number: 9
+  skip: True  # [py==36]
   features:
     - vc9     # [win and py27]
     - vc10    # [win and py34]
-    - vc14    # [win and py35]
+    - vc14    # [win and py>=35]
```
jakirkham: If we skip Python 3.6 above, do we still need to change this?

recipe/meta.yaml
```diff
@@ -36,13 +37,18 @@ requirements:
...
     - libgfortran  # [linux]
+    - vc 9  # [win and py27]
+    - vc 10  # [win and py34]
+    - vc 14  # [win and py>=35]
   run:
     - zlib 1.2.*
     - libgfortran  # [not win]
+    - vc 9  # [win and py27]
+    - vc 10  # [win and py34]
+    - vc 14  # [win and py>=35]
```
jakirkham: I've seen this a few places, but I'm a little nervous about it. How will this work? What do you think about this strategy @msarahan?

recipe/meta.yaml
ocefpaf: This is for humans.

Issue #51, opened 2016-10-19
parallel hdf5


minrk:
Is it possible to get a parallel-enabled build of hdf5? Assuming a serial build is still desirable, how should they be separated? Different package (e.g. `hdf5-parallel`) or feature tracking in the hdf5 package (like numpy+blas)?

I'm looking at building fenics packages, and one impediment that has been pointed out is that it needs parallel hdf5 and vtk, but vtk pulls in this serial hdf5.

mikaem:
@minrk I think the best solution would be to create an hdf5 parallel package. There is also the question of which MPI version (mpich/openmpi), so perhaps even a division into hdf5-mpich or hdf5-openmpi are needed? The parallel hdf5 build uses mpicc to compile and requires the —enable-parallel flag. Not sure this can be combined with a serial package. Note that ubuntu has different packages for serial or parallel hdf5, so perhaps it makes perfect sense that conda-forge have the same?

jakirkham:
PR ( https://github.com/conda-forge/hdf5-feedstock/pull/48  ) is an attempt at a thread safe build. Don't think that is related though, but feel free to correct me.

jakirkham:
I think the big question to answer is how MPI is done. Thus far PR ( https://github.com/conda-forge/staged-recipes/pull/1501 ) of having an MPI feature package seems sensible. If we go that route, we can add a branch at this feedstock for the MPI vs the non-MPI versions.

Maybe you have MPI preferences @mikaem . If so, could you weigh in on PR ( https://github.com/conda-forge/staged-recipes/pull/1501 ) if you haven't already?

mikaem:
@jakirkham PR #48 is not really related.
I like the mpi metapackage. In reality people will want to use either openmpi or mpich, but who's to say which is better? I prefer mpich because it gives me less headache when it comes to installing Fenics, but that hardly counts as a general opinion. I think the performance of both packages are pretty much the same. So the meta package seems like a good solution. I didn't know there could be different branches of the same package, though. Do you have an example using this approach @jakirkham ?

xmjiao:
Has there been some progress regarding parallel hdf5? I think this is a critical feature to be supported (for fenics). Thanks!

jakirkham:
[...]

PR #57, opened 2017-02-17
Continues #48

ocefpaf:
I also added the thread safe option to Windows.

Closes #48

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -1,17 +1,28 @@
 mkdir build
 cd build
 
-REM set environement variables
+:: Set environment variables.
 set HDF5_EXT_ZLIB=zlib.lib
 
-REM configure step
-cmake -G "NMake Makefiles" -DCMAKE_BUILD_TYPE:STRING=RELEASE -DCMAKE_PREFIX_PATH=%LIBRARY_PREFIX% -DCMAKE_INSTALL_PREFIX:PATH=%LIBRARY_PREFIX% -DHDF5_BUILD_CPP_LIB=ON -DBUILD_SHARED_LIBS:BOOL=ON -DHDF5_BUILD_HL_LIB=ON -DHDF5_BUILD_TOOLS:BOOL=ON -DHDF5_ENABLE_Z_LIB_SUPPORT:BOOL=ON %SRC_DIR%
+:: Configure step.
+cmake -G "NMake Makefiles" ^
+      -D CMAKE_BUILD_TYPE:STRING=RELEASE ^
+      -D CMAKE_PREFIX_PATH=%LIBRARY_PREFIX% ^
+      -D CMAKE_INSTALL_PREFIX:PATH=%LIBRARY_PREFIX% ^
+      -D HDF5_BUILD_CPP_LIB=ON ^
+      -D BUILD_SHARED_LIBS:BOOL=ON ^
+      -D HDF5_BUILD_HL_LIB=ON ^
+      -D HDF5_BUILD_TOOLS:BOOL=ON ^
+      -D HDF5_ENABLE_Z_LIB_SUPPORT:BOOL=ON ^
+      -D HDF5_ENABLE_THREADSAFE=ON ^
+      -D ALLOW_UNSUPPORTED=ON ^
+      %SRC_DIR%
 if errorlevel 1 exit 1
 
-REM Build C libraries and tools
+:: Build C libraries and tools.
 nmake
 if errorlevel 1 exit 1
 
-REM Install step
+:: Install step.
 nmake install
 if errorlevel 1 exit 1
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,15 +1,6 @@
 #!/bin/bash
 
-# FIXME: This is a hack to make sure the environment is activated.
-# The reason this is required is due to the conda-build issue
-# mentioned below.
-#
-# https://github.com/conda/conda-build/issues/910
-#
-source activate "${CONDA_DEFAULT_ENV}"
-
-if [ "$(uname)" == "Darwin" ]
-then
+if [ "$(uname)" == "Darwin" ]; then
     export CXX="${CXX} -stdlib=libc++"
     export LDFLAGS="${LDFLAGS} -Wl,-rpath,$PREFIX/lib"
 fi
@@ -23,8 +14,13 @@ export LIBRARY_PATH="${PREFIX}/lib"
             --enable-cxx \
             --enable-fortran \
             --enable-fortran2003 \
-            --with-default-plugindir="${PREFIX}/lib/hdf5/plugin"
+            --with-default-plugindir="${PREFIX}/lib/hdf5/plugin" \
+            --enable-threadsafe \
+            --enable-production \
+            --enable-unsupported \
+            --with-ssl
 
+#             --disable-static \
 make -j "${CPU_COUNT}"
 make check
 make install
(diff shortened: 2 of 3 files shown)


ocefpaf:
@marqh can you take a look if this implements everything you expect from #48?

jakirkham:
Could someone please explain what this warning means for C++ support? A lack of understanding stopped this before.

https://github.com/conda-forge/hdf5-feedstock/pull/48#issuecomment-253822517

jakirkham:
To be totally clear, breaking C++ support is a blocking issue for me. So if this is not breaking C++ support I'm ok with it, but that needs to be demonstrated and/or backed up somehow.

ocefpaf:
> Could someone please explain what this warning means for C++ support? 

LMGTFY: https://github.com/Homebrew/homebrew-science/issues/1888

jakirkham:
Not exactly clear there as it seems to be speculation by others. Though this [response]( https://lists.hdfgroup.org/pipermail/hdf-forum_lists.hdfgroup.org/2009-June/000135.html ) is pretty clear.

TL;DR One can build HDF5 with C++ and Fortran wrappers and threading support enabled. However, only the C API will be thread safe not the C++ or Fortran APIs.

jakirkham:
Also note that the Homebrew developers reverted `--enable-unsupported` because it broke Fortran as noted in issue ( https://github.com/Homebrew/homebrew-science/issues/3054 ). See commit ( https://github.com/Homebrew/homebrew-science/commit/936cca5664c71277184eaec01f5bf315ab3e5527 ) for the reversion.

jakirkham:
Maybe a solution would be to make thread support a feature and add a corresponding metapackage to track that. This way we could build HDF5 without the unsafe APIs (Fortran and C++) when threading is enabled and build Fortran and C++ when threading is disabled. Would this work for you @marqh?

marqh:
Hi @jakirkham @ocefpaf 

my current requirement is for a C API HDF5 library which is thread safe.  I do not have a requirement for C++ or Fortran support for conda-forge provided packages

if a way could be agreed upon to meet this requirement then I am happy to contribute to its delivery

thank you
mark

ocefpaf:
> . I do not have a requirement for C++ or Fortran support for conda-forge provided packages

@astrofrog can you perform some test on the Fortran side to see if this PR is OK. I do have a colleague that uses the package exactly like this PR with both  thread safe and Fortran without any issues. (On Linux... Maybe the homebrew problem is a OS X thing.)

jakirkham:
Hi @qkoziol, was reading your statement about thread safety in HDF5 several years back. Am curious to know if enabling this will still cause issues for C++ and Fortran or if things have changed. Can you please advise us on whether this is an ok way to build HDF5 or not? Will this cause issues for applications using the C++ or Fortran API that do not necessarily lock before calling the API? Thanks.

qkoziol:
I'm fairly confident that you won't have any problems with C++ and threadsafe mode in HDF5, but the situation is still the same as it was in the past: unsupported.

jakirkham:
Thanks @qkoziol.

jakirkham:
@ukoethe, do you think this will cause any problems with VIGRA's HDF5 support?

jakirkham:
cc @stuarteberg

ukoethe:
@jakirkham Why do you suspect that this might interfere with VIGRA's build process?

jakirkham:
It's not a concern with VIGRA's build process. It is a concern w.r.t. threading. In particular would enabling thread support in HDF5 have any unintended consequences with VIGRA? As I'm not really familiar with what goes on in those code paths, I was hoping you would know more.

ukoethe:
Class `ChunkedArrayHDF5` explicitly synchronizes concurrent access to the dataset it references. However, multiple chunked array objects referring to different datasets in the same HDF5 file are not jointly synchronized. Class `HDF5File` and the functionality around it (e.g. random forest I/O) are not synchronized.

I suppose that thread-safe HDF5 adds synchronization of its own. It shouldn't interfere with VIGRA per se (unless it is incompatible with `std::thread`), but I can imagine that naive use of VIGRA's primitives in a multi-threaded programm might lead to deadlock.

What else do you think could go wrong?

jakirkham:
> I suppose that thread-safe HDF5 adds synchronization of its own.

IIUC there is no synchronization around the Fortran or C++ APIs. Though I could be missing something.

ref: https://lists.hdfgroup.org/pipermail/hdf-forum_lists.hdfgroup.org/2009-June/000135.html

ukoethe:
VIGRA builds its own C++ wrappers solely from the C-API, so it might actually work.

qkoziol:
If you have thread-safe C++ wrappers built around the HDF5 C API, you will be fine.

ukoethe:
As I said, VIGRA's `HDF5File` and `HDF5Handle` are not thread-safe, but there is also no reason to share objects of these classes among threads -- I'm not aware of any use-cases where thread-local storage wouldn't work. The only class that may need attention is `HDF5HandleShared`, which should be reimplemented as a wrapper around `std::shared_ptr` and the corresponding atomics to become thread-safe.

marqh:
it is sounding to me like there is a fair amount of support for providing the HDF library built in this way, `threadsafe` and `unsupported`

@jakirkham do you think this represents a stable enough approach to be adopted?  It would make a significant impact for us if it could

It seems clearner to me than trying to manage both non-threadsafe and threadsafe build varients, via metapackages or some other mechanism.

marqh:
Hello @jakirkham 

please may we have an update from you on your thoughts on this work?

Do you still have fundamental concerns about this approach?
Do you have alternative proposals which could be implemented?

thank you
mark

jakirkham:
I remain uneasy about integrating this into the default HDF5 build. It is too difficult to reason about all the effects of this on the many libraries in conda-forge that rely on HDF5. Also I'm not sure how this impacts other things like MPI. ( https://github.com/conda-forge/hdf5-feedstock/issues/51 ) Would much prefer to see this as some separate feature, variant, or similar that a user can optionally enable, but is by default disabled.

minrk:
From the linked HDF5 list comment, this seems okay to me, because:

1. threadsafety only applies to the C API
2. C++ / FORTRAN APIs are unaffected

The argument for `--enable-unsupported` seems to be to prevent the confusion of a user typing:

    ./configure --enable-threadsafe --enable-cxx

and expecting threadsafe C++, which they won't get. It doesn't appear that C++ or FORTRAN is *harmed* by the presence of threadsafe C, just that the flags don't communicate that `threadsafe` only applies to the C API.

If that's correct, the only argument I see against threadsafety would be if there's a significant performance cost in the single-threaded case.

ocefpaf:
It seems we have more people pro than con for this PR. Merging this then. If we find any problem with it we can pull the binaries. (But I really doubt it. I've been using this version of HDF5 for almost an year now.)

cpelley:
Thanks everyone, this helps us out a lot :)

marqh:
@ocefpaf @jakirkham 

this represents a large step forward for us.  This has been the major blocker for my community in adopting conda-forge

I have created #62 requesting that I be adopted as a maintainer, I hope this is reasonable request and that I can help with the ongoing maintenance of this package

ocefpaf:
> This has been the major blocker for my community in adopting conda-forge

I am sorry for that. I've been using this modification for so long in our internal build that I forgot about the `conda-forge` community. I apologize for not pushing this earlier.

PR #68, opened 2017-05-03
Bump version to 1.8.18 to resolve multiple CVEs

qwhelan:
A few CVEs were reported and fixed in hdf5 back in November 2016: http://blog.talosintelligence.com/2016/11/hdf5-vulns.html

The hdf5 release notes for that release are available here: https://support.hdfgroup.org/ftp/HDF5/current18/src/hdf5-1.8.18-RELEASE.txt

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.8.17" %}
+{% set version = "1.8.18" %}
 
 {% set maj_min_ver = ".".join(version.split(".")[:2]) %}
 
@@ -10,7 +10,7 @@ package:
 source:
   fn: hdf5-{{ version }}.tar.gz
   url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-{{ maj_min_ver }}/hdf5-{{ version }}/src/hdf5-{{ version }}.tar.gz
-  md5: 7d572f8f3b798a628b8245af0391a0ca
+  md5: dd2148b740713ca0295442ec683d7b1c
   patches:
     # Patches the test suite to skip the `cache` test.
     # This test has been found to rather resource intensive.
@@ -25,7 +25,7 @@ source:
     - osx_configure.patch  # [osx]
 
 build:
-  number: 11
+  number: 0
   skip: True  # [win and py36]
   features:
     - vc9     # [win and py27]


qwhelan:
Just updated the PR, so believe this is ready for merging once builds complete.

recipe/meta.yaml
```diff
@@ -20,7 +20,7 @@ source:
     - test_Makefile.in.patch
 
 build:
-  number: 10
+  number: 1
```
jakirkham: Please reset to 0.

recipe/meta.yaml
```diff
@@ -20,7 +20,7 @@ source:
...
   skip: True  # [py==36]
```
jakirkham: Please change `py==36` to `win and py==36` and then [re-render]( https://github.com/conda-forge/conda-smithy#re-rendering-an-existing-feedstock ) with the latest `conda-smithy`.

recipe/build.sh
```diff
@@ -3,6 +3,9 @@
 if [ "$(uname)" == "Darwin" ]; then
     export CXX="${CXX} -stdlib=libc++"
     export LDFLAGS="${LDFLAGS} -Wl,-rpath,$PREFIX/lib"
+else
+    # Shared Fortran libs on OSX unsupported as of 1.8.18
+    FORTRAN_FLAGS=("--enable-fortran" "--enable-fortran2003")
```
jakirkham: Hmm...this may be a problem. @astrofrog, care to comment?

qwhelan: @jakirkham While digging into why the Fortran tests were failing on OSX, I came across a log message saying the following:

    checking if shared Fortran libraries are supported... no
    configure: WARNING: Shared Fortran libraries not currently supported on Mac.
    configure: WARNING: Allowing unsupported Fortran shared libraries due to use of --enable-unsupported flag

The ```--enable-unsupported``` flag is needed to build the high-level API and enable thread-safe (which is *not* supported by the library authors), so we can't turn that off.

qwhelan: Also, to clarify - the lack of support for shared fortran libs on OSX goes back to at least 1.8.14.

jakirkham: I see. Thanks for clarifying. What if we do a hack and try to delete the shared library on OS X instead?

qwhelan: I don't have a Mac, so I've been testing through Travis here - my impression is that Apple ships those with OSX. We could try to delete the shared libraries that are built, but we'd need to delete between ```make``` and ```make check```.

Given the fortran test suite is totally failing on OSX, I'd be skeptical of whether it's safe to rely on.

jakirkham: No we use a `gcc` package to handle that case. Considering this built fine before the `--enable-unsupported` flag was added and it is complaining about shared libraries on Mac, I think it would be worth trying to see if forcing static libraries would work. Would recommend borrowing this [commit]( https://github.com/conda-forge/python-feedstock/commit/f2537c24f16c9f92d10b7584e801f2293acdc74e ) and changing `libbz2` to `libgfortran`.

qwhelan: @jakirkham Just passing the ```--disable-shared``` flag seems to have done the trick. Would you like me to throw in something like ```libbz2``` to nuke the old shared lib when users upgrade?

jakirkham: Nice! Thanks for figuring that out. I think removing `libgfortran` is unnecessary then as we know it will fail if it does that.

jakirkham:
Could you please add a separate PR that includes just those changes to fix Mac (e.g. re-rendering, `skip` changes, and the `--disable-shared` flag) for the current version? Also please document in there info about what we learned with Mac.

Once that is merged we can clean up this PR to be a simple version bump and merge it as well.

Does that all sound good?

qwhelan:
@jakirkham Sounds good, splitting it out now

jakirkham:
Thanks for doing that. For reference, this is the split off PR ( https://github.com/conda-forge/hdf5-feedstock/pull/69 ).

jakirkham:
Now that PR ( https://github.com/conda-forge/hdf5-feedstock/pull/69 ) is in. Could we please rebase this?

recipe/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
 
 package:
   name: hdf5
   version: {{ version }}
 
 source:
   fn: hdf5-{{version}}.tar.gz
-  url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-{{ version }}/src/hdf5-{{ version }}.tar.gz
-  md5: 7d572f8f3b798a628b8245af0391a0ca
+  url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8/hdf5-{{ version }}/src/hdf5-{{ version }}.tar.gz
```
jakirkham: Pulling out this URL tweak into PR ( https://github.com/conda-forge/hdf5-feedstock/pull/70 ). So this will be a bit cleaner to manage (e.g. not hard coding version numbers in the URL).

jakirkham: Have addressed this in `master`. So will need another rebase for this PR to clean this stuff out.

qwhelan: Thanks for making the URL change. Rebased and pushed a minute ago.

qwhelan:
@jakirkham I believe this is ready to merge.

qwhelan:
@jakirkham Thanks - you'll need to merge as I don't have rights to.

ocefpaf:
@qwhelan thanks for the PR and sorry for the trouble you had with it.

qwhelan:
@ocefpaf Thanks! And it was my first ```conda-forge``` contribution, so some learning curve was expected.

PR #72, opened 2017-06-11
Build hdf5 with parallel support

nehaljwani:


jakirkham:
There is some discussion of how to approach this for HDF5 in issue ( https://github.com/conda-forge/hdf5-feedstock/issues/51 ) and more generally for conda-forge ( https://github.com/conda-forge/staged-recipes/pull/1501 ). It may even require a new `conda` feature to be totally correct. ( https://github.com/conda/conda/issues/608 ) Though we might be able to get most of the way their with the `staged-recipes` proposal and proceed with that to satisfy this demand. Let's see what comes out of those related discussions so that we know how to implement this correctly.

cc @minrk

nehaljwani:
Thank you for the links. I am closing this PR as I was using it only to test the build.

PR #73, opened 2017-06-13
update_to_1.10.1

ocefpaf:
We should not updated the pinning yet! But I need this version in the channel to build the software against it and run some tests.


jakirkham:
Sure, seem fine to me.

PR #89, opened 2018-06-29
Update to use conda build 3

marcelotrevisani:


marcelotrevisani:
@conda-forge-admin , please rerender

marcelotrevisani:
This PR is just to try to move this recipe to conda build 3

marcelotrevisani:
cc: @ocefpaf @jakirkham

recipe/meta.yaml
```diff
@@ -20,34 +20,27 @@ source:
...
 requirements:
   build:
-    - toolchain
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('fortran') }}
+    - cmake >=3.1
+  host:
     - python         # [win]
     - libtool        # [unix]
```
minrk: libtool should also move to a build dependency

minrk:
I made some comments that should get this closer, but I think this cannot proceed until toolchain_fort's run_exports problem is fixed: https://github.com/conda-forge/toolchain-feedstock/pull/39

minrk:
@marcelotrevisani are you still working on this, or would you like me or someone to have a go at finishing it? The libgfortran fix has landed in toolchain, so should be adopted as soon as the next build is triggered.

marcelotrevisani:
Hi @minrk , I'm kinda a busy these days because of my work, I think I will be able to change that just in the end of the week. However, feel free to do any modification which you want to. :)

minrk:
@marcelotrevisani I added one more commit to this in #91 and I think it'll work.

PR #91, opened 2018-07-04
finish up conda-build 3

minrk:
one extra commit and rerender on top of #89

With toolchain-2.1.6, I think the fortran builds will work now

includes #89 
closes #88


recipe/meta.yaml
```diff
@@ -20,33 +20,25 @@ source:
...
+  number: 1
 
 requirements:
   build:
-    - toolchain
-    - python         # [win]
-    - libtool        # [unix]
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('fortran') }}  # [unix]
```
jakirkham: FWIW `conda-build` doesn't respect this for some strange reason. ( https://github.com/conda/conda-build/issues/2958 )

recipe/meta.yaml
```diff
@@ -98,18 +90,15 @@ test:
...
-
     # Inspect linkages of HDF5.
-    {% for each_hdf5_lib in hdf5_libs %}
     - otool -L $PREFIX/lib/lib{{ each_hdf5_lib }}.dylib                      # [osx]
     {% endfor %}
-    - conda inspect linkages -p $PREFIX $PKG_NAME  # [not win]
-    - conda inspect objects -p $PREFIX $PKG_NAME  # [osx]
+
+    - conda inspect linkages -p $PREFIX $PKG_NAME  # [unix]
+    - conda inspect objects -p $PREFIX $PKG_NAME   # [osx]
```
jakirkham: These can be dropped. `conda-build` 3 already outputs this info and these commands are deprecated.

recipe/meta.yaml
minrk: It's doing the right thing for these, though, right? One of the builds fails if the fortran compiler is selected on Windows because flang doesn't install with vc9.

recipe/meta.yaml
minrk: 👍

minrk:
Build's still working, but timing out on Windows due to the LTCG issue. @isuruf can I ask for you help passing the right flag in bld.bat?

Are the new cb3 compilers on Windows not meant to be the same as those used for pre-cb3 vc9 and vc14 builds? I thought that changing compilers wasn't supposed to be happening yet, but it seems to be compiling with different configurations with cb3 on Windows than before.

minrk:
Thanks @isuruf!  From the build output, it looks like the LTCG flag didn't have the desired effect. The warnings are still there and still causing the build to take an hour instead of 10 minutes. Any ideas?

PR #98, opened 2018-12-07
prioritize nompi via build number

minrk:
instead of deprioritising mpi builds with track_features.

#90 *almost* works great. Deprioritizing via track_features works well as long as a given package has always depended on the given variant, or when that package has a variant scheme that *matches* that of hdf5.

However, there is a case where it doesn't work, and it just so happens to be the exact case that motivated me to build the parallel variants in the first place (fenics). And that's the case where an existing package has been building with serial hdf5 and wants to switch to parallel hdf5. The track_features deprioritization has a high enough priority that if a downstream package has ever been built against serial hdf5, new builds (and new versions) will never be installed unless parallel hdf5 is explicitly requested by the user. I discovered this in fenics, which builds and works great with parallel hdf5 now (yay!) but `conda install fenics` is getting the last version before parallel hdf5. `conda install fenics hdf5=*=mpi*` installs the latest build with parallel hdf5. If any other package has resulted in hdf5 with mpi, feincs with parallel hdf5 will be installed.

build number offsets, as we are using for new-compiler builds, seem to work in a more expected way, priority-wise for this kind of thing.

Changing this here doesn't affect what any downstream packages should do and doesn't require any rebuilds of packages other than this one. It should just work once these are released.

https://github.com/conda-forge/h5py-feedstock/pull/45 which adopts the same track_features approach here should wait for resolution here before being merged.

Checklist
* [x] Used a fork of the feedstock to propose changes
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,9 +1,13 @@
 {% set version = "1.10.4" %}
 {% set maj_min_ver = ".".join(version.split(".")[:2]) %}
-{% set build = 1003 %}
+{% set build = 1004 %}
 
-{# recipe-lint fails if mpi is undefined #}
+# recipe-lint fails if mpi is undefined
 {% set mpi = mpi or 'nompi' %}
+{% if mpi == "nompi" %}
+# prioritize nompi via build number
+{% set build = build + 100 %}
+{% endif %}
 
 package:
   name: hdf5
@@ -31,11 +35,7 @@ source:
 
 build:
   number: {{ build }}
-  # Use track_features to express a preference for
-  # non-mpi variant
-  {% if mpi != 'nompi' %}
-  track_features:
-    - hdf5_{{ mpi }}
+  {% if mpi != "nompi" %}
   {% set mpi_prefix = "mpi_" + mpi %}
   {% else %}
   {% set mpi_prefix = "nompi" %}


minrk:
There is a small downside to using build numbers: When I make a new build of fenics with parallel hdf5, the resolution of `conda upgrade --all` will be ambiguous:

```
Warning: 2 possible package resolutions (only showing differing packages):
  - conda-forge::fenics-2018.1.0-py37h53f436f_6, local::hdf5-1.10.4-mpi_mpich_h046f12a_1002
  - conda-forge::fenics-2018.1.0-py37h53f436f_5, local::hdf5-1.10.4-nompi_h5598ddc_1102
```

This is a short-term problem, though, as it will go away once there is a new version release of fenics with only parallel hdf5 builds. There is no ambiguity when upgrading fenics explicitly.

I've updated my [notes on mpi variants](https://hackmd.io/ry4uI0thTs2q_b4mAQd_qg) with these results.

rabernat:
Hi @minrk. I just wanted you to know that I recently decided I wanted to use an mpi-compatible h5py and that I wanted to install it from conda forge (rather than building from source). I found all of the different issues and PRs you have created in order to make this happen. I am impressed by your persistence! 😉 Thanks for all your work on this thorny problem.

PR #110, opened 2019-04-28
update to hdf5 1.10.5

MarDiehl:
Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)


MarDiehl:
Seems that I misunderstood the patch for t_pflush2 and openmpi

MarDiehl:
Any comments on

```
In file included from ../src/H5private.h:615,
                 from ../src/H5Defl.c:29:
../src/H5win32defs.h:57:66: error: expected expression before ')' token
 #define HDopen(S,F,...)       _open(S, F | _O_BINARY, __VA_ARGS__)
```
for the windows build with 2008 compilervars?

gillins:
Might be worth trying: https://github.com/stnava/ITKR/issues/24

The comment above the error line is interesting:
```
* Also note that the variadic macro is using a VC++ extension
* where the comma is dropped if nothing is passed to the ellipsis.
```

MarDiehl:
Many thanks, that points into the correct direction. Problems with these kinds of macros seem to be quite common, see https://stackoverflow.com/questions/5588855.

My C skills are definitely not sufficient to work on this issue and I have no windows system to test. I would appreciate if anyone more familiar with these things would look into `H5win32defs.h`. Probably it is just one `#if` for older VS versions

gillins:
Looks like this change may have broken it: https://github.com/live-clones/hdf5/commit/734aebc39538039c6e81db63edd68eb3a2029cd2#diff-9bbd760bbce941dc0f4fb74197eb94a3

I wonder if there would be any harm patching it to the old syntax for `win` and `py27`? Do we still need `py27` support on Windows anyway?

ocefpaf:
> I wonder if there would be any harm patching it to the old syntax for win and py27? Do we still need py27 support on Windows anyway?

Harm? Probably not. Do you really need to? I don't think so. We already dropped support for py27 on Windows in `libnetcdf`, `gdal` and many other packages in the stack.

gillins:

PR #112, opened 2019-07-07
Remove static lib

ocefpaf:
I'm planning on sending PRs like these to many of our libraries to reduce the package size and improve cloud deployment experience. 99.99% of the time we don't need the `static` libs, but if we recognize places where we need them we must figure out a way to use multiple `outputs` correctly and make a `static` lib subpackage.


ocefpaf:
I'm not sure the failures on OS X here are related to this PR :-/

```
Fortran API:  fortranlib_test  Test Log
============================
dyld: Library not loaded: $PREFIX/lib/libhdf5.103.dylib
  Referenced from: $SRC_DIR/fortran/test/./fortranlib_test
  Reason: image not found
Command terminated abnormally.
        0.00 real         0.00 user         0.00 sys
make[5]: *** [fortranlib_test.chkexe_] Error 1
make[4]: *** [build-check-s] Error 2
make[3]: *** [test] Error 2
make[2]: *** [check-am] Error 2
make[1]: *** [check-recursive] Error 1
make: *** [check-recursive] Error 1
```

jakirkham:
> I'm planning on sending PRs like these to many of our libraries to reduce the package size and improve cloud deployment experience. 99.99% of the time we don't need the `static` libs...

This seems like a reasonable thing to do.

> ...but if we recognize places where we need them we must figure out a way to use multiple `outputs` correctly and make a `static` lib subpackage.

The one challenge with this is we are unlikely to know which packages use static libraries until we have stripped many of them. So we will have a rather large amount of rebuilding on our hands. I wonder if there is a better way to determine this to mitigate significant future effort. Thoughts?

xhochy:
> I'm planning on sending PRs like these to many of our libraries to reduce the package size and improve cloud deployment experience. 99.99% of the time we don't need the `static` libs, but if we recognize places where we need them we must figure out a way to use multiple `outputs` correctly and make a `static` lib subpackage.

I have the same feeling about static libs and other package size concerns. I will raise a detailed issue later and link it here.

jakirkham:
@xhochy, is arrow statically linking to libraries or dynamically linking?

xhochy:
@jakirkham We moving to dynamic linking on conda slowly. I think there may be 1-2 that still link statically but that is rather historic than intentioned.

xhochy:
Let's continue the general discussion here: https://github.com/conda-forge/conda-forge.github.io/issues/544

xhochy:
@ocefpaf I think the solution could be to run `make install` before `make check`. Then the library would be at the location where the tests expect them.

ocefpaf:
Thanks @xhochy! That fixed it.

@minrk do you foresee any issue with your use of `hdf5` and this PR?

minrk:
I don't foresee any issues here for my use, but definitely +1 to adding `-static` outputs.

ocefpaf:
> I don't foresee any issues here for my use, but definitely +1 to adding `-static` outputs.

The last commit is an attempt to do that. Let's see if that works.

@jjhelmus I based this on your PR to `openssl`. Do you mind taking a look?

recipe/install-static.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+mkdir -p ${PREFIX}/lib
+cp \
```
ocefpaf: Not sure where these files live at this point of the build.

@jjhelmus what was your approach with `openssl`?

jjhelmus: Each output gets a copy of the work directory as it exists after the `build.sh` or `bld.bat` script runs.  The layout of this directory is going to depend on the tools and content of those files.  Adding a `exit 1` to the end of `build.sh` or using `conda debug` will give you chance to examine a copy of the folder.  

You can also examine the Makefile to see what `install` target does, most of the time it just copies various files into the prefix.  Sometimes there is a `make install-static` or similar target which makes the process easier.  

From my examination it looks like the static libraries (and the dynamic ones) are staged in `.libs` directories is various locations:

```
$ find . -name "libhdf*.a"
./c++/src/.libs/libhdf5_cpp.a
./fortran/src/.libs/libhdf5_fortran.a
./hl/c++/src/.libs/libhdf5_hl_cpp.a
./hl/fortran/src/.libs/libhdf5hl_fortran.a
./hl/src/.libs/libhdf5_hl.a
./src/.libs/libhdf5.a
```

jjhelmus:
This layout looks reasonable, although I will propose an alternative that can be considered.

One question that should be resolved if we are going to start splitting out static libraries into their own packages is where the header files should be packaged? When building downstream packages both the dynamic and static libraries require the headers, so ideally they would be available when either is installed.  To prevent clobbering between that static and dynamic packages these files could be placed into their own packages, say `[lib]hdf5-dev` although this will complicated the packages some.  

If splitting the headers into their own package is desired it may be reasonable to follow the pattern used in the [AnacondaRecipes openblas recipe](https://github.com/AnacondaRecipes/openblas-feedstock/blob/ae5af5ed05e5dfe7d626c5d2ce84809900f06764/recipe/meta.yaml).  Here the top level build script (build.sh or bld.bat) both compiles the libraries and installs them into the prefix.  The outputs specify explicitly the files which they contain.

recipe/meta.yaml
```diff
@@ -76,78 +72,115 @@ requirements:
...
-  files:
-    - h5_cmprss.c
-    - h5tutr_cmprss.cpp
-    - h5_cmprss.f90
-    - compound_fortran2003.f90
+outputs:
+  - name: hdf5
+    build:
+      # run_exports:
+      #   - {{ pin_subpackage('hdf5', max_pin='x.x.x') }} {{ build_pin }}
```
jjhelmus: A run_export should be declared

ocefpaf: Ah. I removed this to test it. I cannot build or re-render with a `run_exports` in this line. Not sure what is going on. I guess it is the non-standard build string.

recipe/install-static.sh
ocefpaf: > Adding a exit 1 to the end of build.sh or using conda debug will give you chance to examine a copy of the folder.

Thanks!

>  Sometimes there is a make install-static or similar target which makes the process easier.

I'll check if that is the case here.

ocefpaf:
> One question that should be resolved if we are going to start splitting out static libraries into their own packages is where the header files should be packaged? When building downstream packages both the dynamic and static libraries require the headers, so ideally they would be available when either is installed. To prevent clobbering between that static and dynamic packages these files could be placed into their own packages, say `[lib]hdf5-dev` although this will complicated the packages some.

To be honest I would like to avoid having a `dev` package if we can. While it makes sense and is probably the best approach, as you said, the outputs recipe would get too complicated.

Maybe the `<package>-static` should depend on the `<package>` and the latter would have all the headers, etc. That is not the most optimal option but would avoid having the headers in both packages while still having the headers when installing the static ones.

I don't think we have a case for having only the static libraries, right?

> If splitting the headers into their own package is desired it may be reasonable to follow the pattern used in the [AnacondaRecipes openblas recipe](https://github.com/AnacondaRecipes/openblas-feedstock/blob/ae5af5ed05e5dfe7d626c5d2ce84809900f06764/recipe/meta.yaml). Here the top level build script (build.sh or bld.bat) both compiles the libraries and installs them into the prefix. The outputs specify explicitly the files which they contain.

I like that approach but I'd rather if we could isolate more the built packages so we could use globs instead of listing all the files. I tried that but with globs we end up packaging other stuff that should not be there.

recipe/meta.yaml
ocefpaf: Things worked now. Not sure if I had a <tab> somwhere or if the new conda-build in my env fixed it.

recipe/install-static.sh
ocefpaf: > Sometimes there is a make install-static or similar target which makes the process easier.

I could not find anything, keeping the current approach.

recipe/install.sh
```diff
@@ -0,0 +1,19 @@
...
+if [[ ! ${HOST} =~ .*powerpc64le.* ]]; then
+  # https://github.com/h5py/h5py/issues/817
+  # https://forum.hdfgroup.org/t/hdf5-1-10-long-double-conversions-tests-failed-in-ppc64le/4077
+  make check RUNPARALLEL="${RECIPE_DIR}/mpiexec.sh -n 2"
+fi
+
+rm -rf ${PREFIX}/share/hdf5_examples
+
+# remove the static libraries
+rm -f ${PREFIX}/libs/{libhdf5_cpp,libhdf5_fortran,libhdf5_hl_cpp,libhdf5hl_fortran,libhdf5_hl,libhdf5}.a
```
ocefpaf: This is not working and the `static` libs are making into the package :-/
I'm trying to debug this at the moment.

ocefpaf: Argh! There is an `s` there :-/

ocefpaf:
The packages sizes are:

```
3.2M hdf5-1.10.5-h3c11f04_1101.tar.bz2
1.9M hdf5-static-1.10.5-nompi_h1234567_1101.tar.bz2
```

Hopefully this will improve the situation when deploying `hdf5` on the cloud.

@jjhelmus can you take a second look? What is your opinion on https://github.com/conda-forge/hdf5-feedstock/pull/112#issuecomment-521612166?

jjhelmus:
> I don't think we have a case for having only the static libraries, right?

The use case I have in mind for static libraries is for packages where the risk of getting the wrong dynamic library has **very bad** consequences.  A specific example where statically linking might make sense is `conda-package-handling`.  If that package links is installed with a build of `libarchive` that is missing zstd support or is in anyway broken, conda fails and the user and left with no simply manner of fixing their install.  Currently `conda-package-handling` uses `libarchive` via `py-libarchive-c` but I have a Python extension that directly links to `libarchive` that will be added soon.  In cases such as this I think stability outweighs the benefits of dynamic linking.

If the package with the static libraries requires the non-static version, either explicitly or because of the need for headers, ensuring static linking to the target object is a bit convoluted.  The build script either needs to remove the dynamic libraries  or provide a `-static` or equivalent argument to the linker.  If only the static libraries were installed in the build environment the linker would link those automatically.  Given that static linking is discouraged in conda and I suspect the static package will be used rarely it probably fine (or maybe even desired) to make this process difficult.

jjhelmus:
> Maybe the `<package>-static` should depend on the `<package>` and the latter would have all the headers, etc. That is not the most optimal option but would avoid having the headers in both packages while still having the headers when installing the static ones.

I agree, having `<package>-static` depend on `<package>` ensures the correct header are installed.  Advanced users who can manually delete or hind the dynamic libraries and other portions `<package>` if needed.

ocefpaf:
> I agree, having `<package>-static` depend on `<package>` ensures the correct header are installed. Advanced users who can manually delete or hind the dynamic libraries and other portions `<package>` if needed.

Is that safe enough to avoid the case you made in https://github.com/conda-forge/hdf5-feedstock/pull/112#issuecomment-521851068 ? If so, are you OK merging this one?

ocefpaf:
Let's give this a try. If this breaks thing we can pull the package. I'll copy the model to split static libraries from this one to `libnetcdf`.

beckermr:
@ocepaf This appears broken? I cannot install something simple:

```
localhost:~ Matt$ conda create -n test "hdf5=1.10.5" "libgfortran>3.0.1"
Collecting package metadata (current_repodata.json): done
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: - 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                                                                                                                  
                                                                                                                                                                                                        
UnsatisfiableError: The following specifications were found to be incompatible with each other:                                                                                                         



Package libgfortran conflicts for:
hdf5=1.10.5 -> openmpi[version='>=3.1,<3.2.0a0'] -> libgfortran[version='>=3.0,<4.0.0.a0']
libgfortran[version='>3.0.1']
hdf5=1.10.5 -> libgfortran[version='>=3.0.1,<4.0.0.a0']
```

beckermr:
It appears you did not bump the build number and the packages do not have the mpi prefixes in the build strings.

ocefpaf:
> It appears you did not bump the build number and the packages do not have the mpi prefixes in the build strings.

Ah. Bad merge! The bump I had was not enough after #114. Thanks for #115

PR #122, opened 2020-07-18
Update to version 1.12

djhoese:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Version 1.12 wasn't picked up for some reason so this is my manual update. I've never even built HDF5 myself, but want to play with this version so let's see how this goes. I'd like to try adding `--enable-ros3-vfd` if it isn't auto-added, but I'm not sure where it needs to go.

Closes #121

PR #124, opened 2020-08-04
Add --enable-ros3-vfd build flag

djhoese:
The whole point of #122 was to play with the new read-only S3 VFD functionality of HDF5 1.12. Apparently I forgot to check if it actually got enabled by default (it doesn't). This PR enables the feature and as far as I know does not have any additional requirements or effect any other features of the library since it should be a libcurl-based contact to a remote S3 storage. More info here:

https://portal.hdfgroup.org/display/HDF5/Configuration+and+Setup+for+HDF5+Read+Only+S3+VFD

Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -37,7 +37,7 @@ fi
             --enable-unsupported \
             --enable-using-memchecker \
             --enable-static=yes \
-            --enable-ros3-vfd \
+            --enable-ros3-vfd
 
 # allow oversubscribing with openmpi in make check
 export OMPI_MCA_rmaps_base_oversubscribe=yes
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -42,6 +42,9 @@ fi
 # allow oversubscribing with openmpi in make check
 export OMPI_MCA_rmaps_base_oversubscribe=yes
 
+if [[ "$CI" != "travis" ]]; then
+  make -j "${CPU_COUNT}" ${VERBOSE_AT}
+else
 # using || to quiet logs unless there is an issue
 {
     # see this https://github.com/travis-ci/travis-ci/issues/4190#issuecomment-353342526
@@ -55,4 +58,4 @@ export OMPI_MCA_rmaps_base_oversubscribe=yes
     tail -n 5000 make_logs.txt
     exit 1
 }
-
+fi
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -37,6 +37,7 @@ fi
             --enable-unsupported \
             --enable-using-memchecker \
             --enable-static=yes \
+            --enable-ros3-vfd \
 
 # allow oversubscribing with openmpi in make check
 export OMPI_MCA_rmaps_base_oversubscribe=yes
--- a/.ci_support/linux_64_mpimpich.yaml
+++ b/.ci_support/linux_64_mpimpich.yaml
@@ -16,9 +16,17 @@ fortran_compiler:
 - gfortran
 fortran_compiler_version:
 - '7'
+libcurl:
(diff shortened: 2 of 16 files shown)


recipe/build.sh
```diff
@@ -37,6 +37,7 @@ fi
             --enable-unsupported \
             --enable-using-memchecker \
             --enable-static=yes \
+            --enable-ros3-vfd \
```
scopatz, line 40: last element shouldn't have a trailing slash

djhoese: I thought it was odd but the static one had it too. Any idea why that works?

djhoese:
```
checking if the Read-Only S3 virtual file driver (VFD) is enabled... no
configure: error: The Read-Only S3 VFD was requested but cannot be built.
                      Please check that openssl and cURL are available on your
                      system, and/or re-configure without option
                      --enable-ros3-vfd.
```

Looks like SSL and curl are needed. For some reason I assumed they would have already been added to the conda-forge build. I'm guessing that makes this a bigger deal to include (people maybe not wanting ssl and curl installed)?

recipe/build.sh
scopatz: Probably because there are just blank lines following it

scopatz:
I mean, I don't think it is that big of a deal to add ssl and curl deps

djhoese:
As a loss again. Configure seems to find the headers for curl just fine:

```
2020-08-06T20:56:17.8136018Z checking curl/curl.h usability... yes
2020-08-06T20:56:17.8485835Z checking curl/curl.h presence... yes
2020-08-06T20:56:17.8490187Z checking for curl/curl.h... yes
```

But then later during compilation:

```
2020-08-06T21:05:52.0771711Z Making install in src
2020-08-06T21:05:52.4242117Z make[1]: Entering directory '/home/conda/feedstock_root/build_artifacts/hdf5_split_1596747167656/work/src'
2020-08-06T21:05:52.4243388Z   CC       H5FDros3.lo
2020-08-06T21:05:52.9916273Z In file included from H5FDros3.c:35:0:
2020-08-06T21:05:52.9917163Z H5FDs3comms.h:60:10: fatal error: curl/curl.h: No such file or directory
2020-08-06T21:05:52.9917621Z  #include <curl/curl.h>
2020-08-06T21:05:52.9917909Z           ^~~~~~~~~~~~~
2020-08-06T21:05:52.9918201Z compilation terminated.
```

scopatz:
Maybe you need to pass the location of curl into the configure script? Sometimes that is needed

djhoese:
I'm 90% the configure script is the thing that is giving the first output of "checking" and successfully finding curl.h. Otherwise would I do `--with-curl="${PREFIX}/include/curl`?

djhoese:
@isuruf I suppose we should just remove the verbosity stuff and then this is good for review?

isuruf:
Why remove verbose stuff? It's good for debugging when something fails.

djhoese:
Ah ok. I just thought you were adding it to help figure out why curl wasn't showing up. Nevermind then. Ready for review.

isuruf:
> I just thought you were adding it to help figure out why curl wasn't showing up. 

I was. It'll still be useful for future when something goes wrong.

scopatz:
Great work @djhoese and thanks @isuruf!

PR #134, opened 2020-11-17
Fix ros3 driver for curl>7.69 and enable ros3 on Windows

### LINKED ISSUE #125: Read-only S3 driver (ros3) doesn't seem to work
I created #122 and #124 so I could play with the S3 reading ability of HDF5. I tried using a public S3 bucket to inspect a NetCDF4 file with no luck. I then started looking at one of the files here: https://www.hdfgroup.org/solutions/enterprise-support/cloud-amazon-s3-storage-hdf5-connector/ but got the same errors. Here's a command from the example video that should work:

```
h5dump -pBH --filedriver=ros3 "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/snp500.h5"  
```

But just produces:

```
h5dump error: unable to open file "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/snp500.h5"
```

I know this is probably an upstream issue but wanted to make this so if anyone else tries this feature they know it isn't working. Also, is there a way to install the built package from a PR so I could try enabling some debug flags to further debug this issue?

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list
# packages in environment at /home/davidh/miniconda3/envs/hdf5_test:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
c-ares                    1.16.1               h516909a_0    conda-forge
ca-certificates           2020.6.20            hecda079_0    conda-forge
hdf5                      1.12.0          nompi_h54c07f9_101    conda-forge
krb5                      1.17.1               hfafb76e_2    conda-forge
libcurl                   7.71.1               hcdd3856_4    conda-forge
libedit                   3.1.20191231         h46ee950_1    conda-forge
libev                     4.33                 h516909a_0    conda-forge
libgcc-ng                 9.3.0               h24d8f2e_14    conda-forge
libgfortran-ng            7.5.0               hdf63c60_14    conda-forge
libgomp                   9.3.0               h24d8f2e_14    conda-forge
libnghttp2                1.41.0               hab1572f_1    conda-forge
libssh2                   1.9.0                hab1572f_5    conda-forge
libstdcxx-ng              9.3.0               hdf63c60_14    conda-forge
ncurses                   6.2                  he1b5a44_1    conda-forge
openssl                   1.1.1g               h516909a_1    conda-forge
tk                        8.6.10               hed695b0_0    conda-forge
zlib                      1.2.11            h516909a_1007    conda-forge
```
</details>

<br/>
Details about  <code>conda</code> and system ( <code>conda info</code> ):
<details>

```
$ conda info

     active environment : hdf5_test
    active env location : /home/davidh/miniconda3/envs/hdf5_test
            shell level : 2
       user config file : /home/davidh/.condarc
 populated config files : /home/davidh/.condarc
          conda version : 4.8.3
    conda-build version : not installed
         python version : 3.7.3.final.0
       virtual packages : __cuda=10.2
                          __glibc=2.31
       base environment : /home/davidh/miniconda3  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/davidh/miniconda3/pkgs
                          /home/davidh/.conda/pkgs
       envs directories : /home/davidh/miniconda3/envs
                          /home/davidh/.conda/envs
               platform : linux-64
             user-agent : conda/4.8.3 requests/2.23.0 CPython/3.7.3 Linux/5.4.0-7634-generic pop/20.04 glibc/2.31
                UID:GID : 53807:1000
             netrc file : None
           offline mode : False
```
</details>

djhoese:
Fix #125 

See related issue for the reasoning behind this. The general idea is that the "NOBODY" object used for libcurl stops headers from being included, but setting it to NULL or `0` (false) doesn't turn it back on. The patch in this PR re-enables it by setting HTTPGET to 1.

CC @dopplershift

Checklist


djhoese:
@conda-forge-admin please rerender

yarikoptic:
and what if `curl <= 7.69` -- would it still work? (just checking)

djhoese:
That's a question for @dopplershift I think.

yarikoptic:
IMHO might be worth adding some explicit unittest for that, e.g. 
```
h5dump --filedriver=ros3 "http://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | head -n 1 | grep '^HDF5' 
```
or alike (that one seems to be the smallest file in that bucket/directory - just 8k, and I verified that even that one doesn't work without some fix ;))

djhoese:
I've been using:

```
h5dump -pBH --filedriver=ros3 "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/snp500.h5"
```

Which I think I got from their own blog post or documentation on this driver when it was released.

yarikoptic:
I was using that one too locally, but for CI better to use smaller file. I have no opinion about options ;-)

dopplershift:
Based on my reading of the docs, what we’re changing with NOBODY vs. HTTPGET is perfectly fine, regardless of version. No clue whatsoever why it became important for curl >7.69.

yarikoptic:
well, my personal confusion is why is it working for me on debian, or when LD_PRELOADing its libcurl 7.72.0. But oh well -- if works (and tests pass), then "who cares?" (until someone does) ;)

djhoese:
@yarikoptic Thoughts on the failing test since `grep` isn't available on Windows? Skip the test on Windows? Or just hope that h5dump gives a logical exit code if it fails and remove the head/grep part of your command?

dopplershift:
@yarikoptic A fair question. Conda-forge isn't applying any patches, so it's either a difference of config/downstream libraries, or Debian is applying a patch. 🤷‍♂️

yarikoptic:
re test: there is `m2-grep` on windows -- should work? `head` is not needed really - left over when I thought to minimize traffic while testing on larger file (was hoping for early termination, not sure if happened anyways)

djhoese:
Turns out I didn't even have ros3 enabled for Windows. Oops.

djhoese:
@yarikoptic I added the m2-grep dependency (never used this package before) to the tests and made sure to include the ros3 driver in the Windows build and now things seem to be passing on Windows. This might be good to go then.

satra:
@djhoese - are there any specific elements to review or test before this can be merged?

djhoese:
@satra I don't think there is much else we can test. We have the basic test in the recipe to make sure the S3 driver works and there are a lot of other tests for all of the existing functionality. Since the patch here is only in the ros3 part of the source I can't imagine it would effect anything else. Since the ros3 driver is enabled in the existing build (on master) I would imagine any issues it would have caused would be found already.

@conda-forge/hdf5 Any other concerns?

beckermr:
I cannot comment on the ros-specific stuff here. However, do note that this PR will only fix the version on master. If you need a fix for the versions in our global pinnings, then you'll need to port this PR to the v1.10.x branch.

djhoese:
Ah forgot about the pinnings. This ros3 driver is only available in v1.12 anyway so no back porting is possible.

satra:
> This ros3 driver is only available in v1.12 anyway so no back porting is possible.

the ros3 driver is available from 1.10.6 onwards

djhoese:
> the ros3 driver is available from 1.10.6 onwards

Ah ok. Well then someone will need to port my patches and the various flags (which I think changed between 1.10 and 1.12) for enabling ssl/curl/ros3. I won't have time to do it.

beckermr:
Sounds good @djhoese. Thanks for all of the hard work here!

satra:
@djhoese - thanks for getting this going. i would be happy to submit one for the 1.10.6 branch based on this.

djhoese:
@satra It will need to be a combination of this PR and #124. This PR includes the fixes for curl and enabling it for windows (I'll fix the title), but #124 was the original work for this driver. There is obviously another PR for updates to 1.12 where I re-created all the patches but most of that shouldn't be needed.

beckermr:
OK. Well in the absence of other feedback, let's rerender since it has been a week and then we can merge. 

@conda-forge-admin rerender

beckermr:
@djhoese Have you check the box to allow the bot to commit or is this a bug?

djhoese:
image

That's how it has always been (I never uncheck it).

djhoese:
You can even see I did a rerender near the top of this PR. So this worked in the past.

djhoese:
Could this be a problem because I created a fork before the github action bot existed?

djhoese:
@conda-forge-admin please rerender

djhoese:
@beckermr :man_shrugging: 

Works when I do it. Maybe it is because you didn't say "please" :wink:

beckermr:
@conda-forge-admin please rerender

beckermr:
The rerender job ran, but it said the permissions were wrong.

beckermr:
Hopefully, this isn't a sign that rerendering is broken globally. 🤞

PR #141, opened 2020-11-27
Fix/1.10.x ros3

satra:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [X] Reset the build number to `0` (if the version changed)


recipe/meta.yaml
```diff
@@ -1,7 +1,7 @@
-{% set version = hdf5_version|default("1.10.5") %}
+{% set version = "1.10.6" %}
```
beckermr, line 1: Do not change this line. This branch builds both 1.10.5 and 1.10.6

satra:

recipe/meta.yaml
```diff
@@ -1,7 +1,7 @@
...
 
 {% set maj_min_ver = ".".join(version.split(".")[:2]) %}
-{% set build = 1011 %}
+{% set build = 0 %}
```
beckermr, line 4: You need to put the build number to 1012.

beckermr (changes requested):
You'll need to add code to select the flags/patches you want only on the 1.10.6 build.

Alternatively, you can wait until the 1.10.5 to 1.10.6 migration is done and then only build 1.10.6 on this branch.

satra:
this follows up on #134 to enable the patch for 1.10.x branch. on my local machine i checked the diff, and the same patch is applicable there. so i cherrypicked most of @djhoese commits.

recipe/meta.yaml
satra: ah, the ros3 driver is only available from 1.10.6 onwards. how would i address that given the changes.

beckermr: You can use jinja2 if blocks or selectors on the version

satra:
@beckermr - i changed meta.yaml to respect version, but is there an example of checking for the version in build.sh and install.bat? the configure flags need to be adjusted.

recipe/meta.yaml
satra: resolved this in meta, but not in install.bat or build.sh

beckermr:
The same variables in the meta.yaml selectors should be in the bash in build.sh

satra:
@beckermr - now i am a little confused. this line for master enables the ros3 driver, but this driver is only available as of 1.10.6 (see: https://www.hdfgroup.org/2019/12/release-of-hdf5-1-10-6-newsletter-170/)

https://github.com/conda-forge/hdf5-feedstock/blob/1.10.x/recipe/build.sh#L40

i don't know whether i should be fixing this, or simply wait for the tests. i'll have to download 1.10.5 code to check if they already had that driver in there or configure simply ignores the flag.

djhoese:
Looks like someone copied my commits from master and applied them to the 1.10 branch where they may or may not be used. If they don't error out it is hard to tell.

satra:
thanks @djhoese - i have fixed the ros3 check and requirements to be specific to 1.10.6, but indeed the builds seem to be working fine on all platforms.  (there is one error, but that was an http timeout)

beckermr:
If the tests of the functionality pass (and do double check they are executed), then I'd say we are good to go.

satra:
@conda-forge-admin, please restart ci

satra:
@beckermr - this is ready for any final reviews and if all looks good, the ros3 driver tests pass with 1.10.6

Issue #185, opened 2023-01-14
Getting rid of the `-static` builds


hmaarrfk:
### Comment:

These are somewhat annoying to maintain. I have only seem them used in the https://github.com/conda-forge/hyperion-fortran-feedstock which seems unmaintained and likely doesn't need them.

Does anybody have objections to this?

hmaarrfk:
xref: https://github.com/conda-forge/hdf5-feedstock/pull/112

hmaarrfk:
I'm mostly trying to be opportunistic here since the new release of 1.14.0 will cause a migration for us. So it shouldn't break anything that anybody is pinned on earlier versions for.

timsnyder:
> These are somewhat annoying to maintain.

Because you have needed to do a migration for new versions?  Just trying to understand the annoyances.

We have used hdf5-static in environments, not as a dependency for building other conda-forge packages and it would be useful to have them available as something installable by end users.  Would it be horrible to add them back for that purpose?

hmaarrfk:
we don't want to ship the static libraries in addition to everything else, this doubles the storage requirements.

I would accept PRs that follow
https://github.com/conda-forge/cfep/blob/main/cfep-18.md?plain=1

That is: create a split package called `hdf5-static` with the static libraries you need.

hmaarrfk:
I guess i removed the `-static` package.

The real issue is that the maintenance of HDF5 has generally been pretty slow. 

Each version is already difficult and having these "dual builds" is somewhat tricky where you have to essentially do the testing twice....

1.14.4 was only merged 2 weeks ago https://github.com/conda-forge/hdf5-feedstock/pull/231

but was released in May.... https://github.com/HDFGroup/hdf5/releases/tag/hdf5_1.14.4.3


so even the current status is very slow at conda-forge for maintainability.

PR #188, opened 2023-01-14
1.14.0 and remove static builds

hmaarrfk:
I really don't think we should be dragging along the static builds. The build process isn't really nice toward building both the static and the dynamic library. I think we are going to be better off without this for 1.14.0

Let me know what you all think

- [ ] Check if I have to include static builds again https://github.com/conda-forge/hdf5-feedstock/issues/185
- [x] Discuss skipping the ROS3 test on windows -- https://github.com/HDFGroup/hdf5/issues/2406
- [x] Atomicity tests known to fail: https://github.com/HDFGroup/hdf5/issues/2196
- [x] Remove huge patch to rename executable files in cmake in favor of renaming them in the build script

Upstream changes

- Upstream changed from -show to -showconfig for cmake generated h5cc https://github.com/HDFGroup/hdf5/blob/e4fb9aa96ef805e2dada8a78c32aab8cf8c34f4a/release_docs/HISTORY-1_13.txt#L1111
- Patched out the use of the statically linked binaries for windows making the comrpessed package 3x smaller

TODO:
- [ ] Create a branch for 1.12.X maintenance (this will be necessary since HDF5 is typically a hard transition)

Checklist


hmaarrfk:
The really heavy handed patching is to remove all the statically linked binaries from windows:
image

hmaarrfk:
The executables that link to the shared libraries take 10s of kB, those that statically link take MBs
image

hmaarrfk:
@djhoese i'm not sure if you are still interested in ROS3 functionality on windows.

djhoese:
It isn't clear to me what ROS3 has to do with this PR or static builds. ROS3 should be made to work on all platforms if upstream supports it. If the tests fail on Windows due to some CI limitation then OK.

recipe/meta.yaml
```diff
@@ -81,127 +90,84 @@ requirements:
...
+    {% for each_hdf5_lib in hdf5_libs %}
+    - test -f $PREFIX/lib/lib{{ each_hdf5_lib }}${SHLIB_EXT}  # [not win]
+    - if not exist %PREFIX%\\Library\\lib\\{{ each_hdf5_lib }}.lib exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\bin\\{{ each_hdf5_lib }}.dll exit 1  # [win]
+    {% endfor %}
+    # Check ros3 driver
+    - h5dump --filedriver=ros3 "http://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | grep '^HDF5'  # [unix]
+    # Jan 2023 -- hmaarrfk
+    # I am really not sure why the ROS driver test doesn't pass on windows
+    # - h5dump --filedriver=ros3 "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | findstr '^HDF5'  # [win]
```
hmaarrfk, line 170: @djhoese i had to skip the windows test here.

hmaarrfk:
@conda-forge-admin please rerender

recipe/meta.yaml
hmaarrfk: Trying it locally, i also get 
```
h5dump error: unable to create FAPL for file access
```

hmaarrfk: Yeah, i'm not sure, on windows the header has the line uncommented meaning that cmake thinks that it compiled with ROS3 support
```
#define H5_HAVE_ROS3_VFD 1
```

djhoese: I feel like I might be missing some context here. Regardless, is this something that should be filed as a bug/discussion upstream?

hmaarrfk: done! they are on it! See link at the top level post!

mkitti:
Do we need to synchronize this with a h5py update?

h5py 3.8.0 was released a few hours ago:
https://github.com/h5py/h5py/releases/tag/3.8.0

mkitti:
h5py was updated to 3.8.0:
https://github.com/conda-forge/h5py-feedstock/pull/125

This incorporates the patches needed for 1.14 support.

hmaarrfk:
> Do we need to synchronize this with a h5py update?

In essence, not really. h5py is pinned to the previous version of HDF5 through the run_export.

So this could be merged, without any fear of breaking existing user's H5PY installation.

Im mostly leaving this here for people to comment on the removal of the static builds. But to me, it seems that they were added as an abundance of caution.

mkitti:
It would be nice if the shared versions were aliased to the static versions in this case via a symlink or just copying so that typing `h5ls` would still work.

hmaarrfk:
`h5ls-dynamic` seems to be a creation of the `cmake` files.

doesn't h5ls still work? The tests say they do.

I aggressively patched windows to make `h5ls` exist but to remove `h5ls-dynamic`. I'm honestly now proud of the aggressive patching. Maybe we should just "rename" the executable instead of patching so much. I'm happy to report this as an issue upstream as well.

mkitti:
Ah, I see. It might be worth discussing at https://forum.hdfgroup.org/ first?

mkitti:
It might make sense to split the tools off to another package. Some Linux distributions do this:
https://packages.debian.org/bullseye/hdf5-tools
https://pkgs.org/search/?q=hdf5-tools

hmaarrfk:
> discussing at https://forum.hdfgroup.org/ first?

Presently, I am too tired to articulate to a wider audience, but maybe over the weekend i can formulate my thoughts?

I am tempted to reimplement this as a "file rename" in our build script instead of a huge patch.

hmaarrfk:
> It might make sense to split the tools off to another package

Two aspects: 

1. conda doesn't really make this "easy". For example, with RPMs I can specify a list of files I want in a subpackage, then everything else goes in the `main` package. This ensures I don't "miss" anything in the subpackage creation. This process is a little manual, and error prone.
2. conda has historically been more "batteries included" (including more options than "the bare minimum") than other distributions.  The binaries on linux only take up 2MB / 11.1MB as reported by `ncdu`. Unless somebody wants to step up and "help maintain", I'm not so inclined to split the package given `1.`.  As a note, uncompressed, the headers make up 2.8/11.1 MB.

hmaarrfk:
@conda-forge-admin please restart cis

hmaarrfk:
@conda-forge/hdf5 this is ready for review.

I'm hoping to merge Feb 5.

ocefpaf (approved):

recipe/meta.yaml
```diff
@@ -75,133 +80,91 @@ requirements:
...
+    {% for each_hdf5_lib in hdf5_libs %}
+    - test -f $PREFIX/lib/lib{{ each_hdf5_lib }}${SHLIB_EXT}  # [not win]
+    - if not exist %PREFIX%\\Library\\lib\\{{ each_hdf5_lib }}.lib exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\bin\\{{ each_hdf5_lib }}.dll exit 1  # [win]
+    {% endfor %}
+    # Check ros3 driver
+    - h5dump --filedriver=ros3 "http://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | grep '^HDF5'  # [unix]
+    # Jan 2023 -- hmaarrfk
+    # I am really not sure why the ROS driver test doesn't pass on windows
+    - h5dump --filedriver=ros3 --s3-cred="(,,)" "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5"  # [win]
```
hmaarrfk, line 167:
```suggestion
    # Jan 2023 -- hmaarrfk. Tests will fail on windows without --s3-cred specified
    # https://github.com/HDFGroup/hdf5/issues/2406
    - h5dump --filedriver=ros3 --s3-cred="(,,)" "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5"  # [win]
```

recipe/meta.yaml
```diff
@@ -75,133 +80,91 @@ requirements:
...
+    {% for each_hdf5_lib in hdf5_libs %}
+    - test -f $PREFIX/lib/lib{{ each_hdf5_lib }}${SHLIB_EXT}  # [not win]
+    - if not exist %PREFIX%\\Library\\lib\\{{ each_hdf5_lib }}.lib exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\bin\\{{ each_hdf5_lib }}.dll exit 1  # [win]
+    {% endfor %}
+    # Check ros3 driver
+    - h5dump --filedriver=ros3 "http://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | grep '^HDF5'  # [unix]
+    # Jan 2023 -- hmaarrfk. Tests will fail on windows without --s3-cred specified
+    # https://github.com/HDFGroup/hdf5/issues/2406
+    - h5dump --filedriver=ros3 --s3-cred="(,,)" "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | findstr '^HDF5'  # [win]
```
hmaarrfk, line 167:
```suggestion
    - h5dump --filedriver=ros3 --s3-cred="(,,)" "https://s3.amazonaws.com/hdfgroup/data/hdf5demo/tall.h5" | findstr "^HDF5"  # [win]
```

hmaarrfk:
Thank you all for your reviews.

PR #203, opened 2023-06-04
adding fortran support to win version

ldallolio:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [X] Bumped the build number (if the version is unchanged)
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [X] Ensured the license file is being packaged.

### DIFF
--- a/.ci_support/linux_64_mpimpichopenssl1.1.1.yaml
+++ b/.ci_support/linux_64_mpimpichopenssl1.1.1.yaml
@@ -1,7 +1,7 @@
 c_compiler:
 - gcc
 c_compiler_version:
-- '11'
+- '12'
 cdt_name:
 - cos6
 channel_sources:
@@ -11,17 +11,17 @@ channel_targets:
 cxx_compiler:
 - gxx
 cxx_compiler_version:
-- '11'
+- '12'
 docker_image:
 - quay.io/condaforge/linux-anvil-cos7-x86_64
 fortran_compiler:
 - gfortran
 fortran_compiler_version:
-- '11'
+- '12'
 libaec:
 - '1'
 libcurl:
-- '7'
+- '8'
 mpi:
 - mpich
 mpich:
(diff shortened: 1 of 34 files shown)


varlackc (commented):
There seem to be some architectures that are still failing.

ldallolio:
> There seem to be some architectures that are still failing.

Hi, indeed (I marked this PR as a draft), I have multiple issues maybe someone can point me in the right direction :
- I do not know why (maybe because of rerendering?) other architecture than windows should fail : I only touched meta.yaml with [win] specifiers and bld.bat, I will check this later before removing the draft mark
- I tried adding mingw-gfortran as a compiler but cmake will also use it as a linker and the ninja generator will pass link flags for visual studio. I looked around but could not find any way to force another linker (link.exe) or flag options, so I suppose this is a no go
- I tried replacing all the compilers in Windows with mingw but the build fails (I will loook into that)
- I did not try yet using flang for Windows, I will

Any help or hint is appreciated!
All the best,

nicogodet:
> I tried replacing all the compilers in Windows with mingw but the build fails (I will loook into that)

I have a successful build with MinGW compilers 
See recipe here : https://github.com/nicogodet/hdf5-otm-feedstock/tree/main/recipe

hmaarrfk:
I'm not sure it will be abi compatible with the non mingw version.

nicogodet:
One idea would be to add an additionnal package like it's done by netcdf ?
https://github.com/conda-forge/netcdf-fortran-feedstock/tree/main

hmaarrfk:
lets not add a new package. Nicolas, it seems you may have a working package for HDF5 + fortran using m2w64, if you make a PR that can be the start of the review process, otherwise, it is very difficut to review without one.

nicogodet:
Will do it soon ;)

PR #200, opened 2023-04-12
BUG: Add upstream gh-2667 for cmake paths

larsoner:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Used `git format-patch -1 b77cb39` (commit from https://github.com/HDFGroup/hdf5/pull/2667) then removed the release notes change from the diff.

Closes #196

PR #204, opened 2023-06-22
test building against hdf5 with cmake

minrk:
only a test change, not a rebuild, so no build bump needed

this _should_ pass with the changes in 1.14.1 (it ought to fail in 1.14.0) #200


varlackc (commented):
The changes look good, however, there are some small changes that might be a good idea to take a look at in order to improve the recipe. 

Currently, the `home` url seems to redirect to the following location
https://www.hdfgroup.org/solutions/hdf5/

https://github.com/conda-forge/hdf5-feedstock/blob/3bf39fafa0fc3fb0d25e49efca106be10e03d58c/recipe/meta.yaml#L167

It might be a good idea to change the URL to the following 

home: https://www.hdfgroup.org/solutions/hdf5/

minrk:
@varlackc thanks! Updated.

minrk:
build succeeded with cmake, confirming that #200 is fixed. Merging.

PR #218, opened 2024-05-23
windows mpi build with impi

minrk:
draft because who knows if this will work


minrk:
e0dd8b38c59b15b0b9149579fae681b324c16b44 completed successfully, going to try to do some cleanup now.

minrk:
Somewhat surprisingly, seems to work!

ax3l (approved):
Cool!

recipe/meta.yaml
```diff
@@ -74,13 +76,11 @@ requirements:
     # Keep both versions of openssl in sync
     - openssl
   host:
-    - {{ mpi }}  # [mpi != 'nompi']
+    - {{ mpi | replace("impi", "impi-devel") }}  # [mpi != 'nompi']
```
hmaarrfk, line 79: string parsing is "clever" and assumes that `impi` doesn't appear as a substring in other options.

```suggestion
    - {{ mpi }}   # [mpi != 'nompi' or mpi != 'impi']
    - impi-devel  # [mpi == 'impi']
```

hmaarrfk (approved):

PR #227, opened 2024-08-29
update version, with new format

itcarroll:
HDF5Group has introduced a "release number" after a hyphen into their distributions. I don't know what distinguishes this from a new patch number. I think we can accommodate by updating the conda-forge build number in case of a new release number: if HDF5 releases a 1.14.4-4, we can keep the version as 1.14.4 and update the build number.

Checklist
* [ ] ~~Bumped the build number (if the version is unchanged)~~
* [x] Reset the build number to `0` (if the version changed)


ajelenak:
There will not be 1.14.4-4 or any -_n_. This is not a new version format but happened as the CI jobs for making release packages had some problems.

The next release will be 1.14.5. Hope this clarifies the situation.

itcarroll:
Thanks, @ajelenak. Would you be able to add a tarball with just "1.14.4" to https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.14/hdf5-1.14.4/src/? Then there would be no need to change the conda recipe (and then change it back for 1.14.5).

itcarroll:
I'm probably not going to be able to resolve the build errors on the subset of platforms where it fails. If anybody wants to close this PR and open a new one for 1.14.4, please feel free.

ajelenak:
> Would you be able to add a tarball with just "1.14.4" to https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.14/hdf5-1.14.4/src/? Then there would be no need to change the conda recipe (and then change it back for 1.14.5).

I suggest to wait for 1.14.5, which is due in a month or so.

zklaus:
Waiting is fine for me. Regarding the build artifacts (I came here via the discussion on element), @itcarroll, you have to rerender after changing the config.

zklaus:
We can wait, but if you want to work on this, @itcarroll, here are some things that I found so far:

- The patches need updating. They seem to be mostly fine, but there is a bit of an offset and in the case of osx-arm64, particularly for `0001-avoid-test-execution-when-cross-compiling.patch.native`, they are just to big and the patching fails. Perhaps this is simply due to new tests. I would just try to re-create the patch manually.
- For Windows, everything seems to work until we try to [copy `h5repart-shared.exe` to `h5repart.exe`](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1014368&view=logs&j=c8e56b94-24b6-5bce-7779-210e8e22265c&t=8149e0a5-8c1b-5470-1341-008d8b4386a3&l=4965). However, there is already `h5repart.exe`, so maybe we just don't need to copy? In other words, I think we can simply remove everything beyond [line 72](https://github.com/conda-forge/hdf5-feedstock/blob/e37dd942f8948c7a725fbb5975f1a13b96d02b96/recipe/bld.bat#L72) from our `bld.bat`.

@ajelenak, can you confirm that the windows cmake build was changed to not add `-shared` to every exe?

If this is not confirmed, we should double check that we indeed still avoid static linking...

The only linux failure so far is a fluke, that will be resolved by restarting.

ajelenak:
> @ajelenak, can you confirm that the windows cmake build was changed to not add `-shared` to every exe?

@zklaus Yes, see: https://github.com/HDFGroup/hdf5/blob/hdf5_1.14.4.3/release_docs/RELEASE.txt#L162-L168.

PR #234, opened 2024-10-22
[DEBUG] Add test for C++ exception error

xylar:
This is a test PR to see whether the test that @gillins added in #233 passes for 1.14.4

Checklist

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -160,7 +160,7 @@ if [[ ("$target_platform" != "linux-ppc64le") && \
 fi
 
 # test for hdf5 C++ exceptions
-$CXX $RECIPE_DIR/testhdf5exc.cpp -o testhdf5exc -L$PREFIX/lib -lhdf5_cpp -lhdf5 -I$PREFIX/include
-LD_LIBRARY_PATH=$PREFIX/lib DYLD_LIBRARY_PATH=$PREFIX/lib ./testhdf5exc
-
-
+if [[ "${CONDA_BUILD_CROSS_COMPILATION:-}" != "1" || "${CROSSCOMPILING_EMULATOR}" != "" ]]; then
+    $CXX $RECIPE_DIR/testhdf5exc.cpp -o testhdf5exc -L$PREFIX/lib -lhdf5_cpp -lhdf5 -I$PREFIX/include
+    ./testhdf5exc
+fi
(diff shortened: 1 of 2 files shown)


xylar:
@gillins, osx-64 is passing here, too.  So I don't think we have an explanation for why kealib is failing in https://github.com/conda-forge/kealib-feedstock/pull/65

gillins:
Sigh. Thanks @xylar. @petebunting any chance you'll have some time to look into this some more?

recipe/build.sh
```diff
@@ -158,3 +158,9 @@ if [[ ("$target_platform" != "linux-ppc64le") && \
   # https://forum.hdfgroup.org/t/hdf5-1-10-long-double-conversions-tests-failed-in-ppc64le/4077
   make check RUNPARALLEL="mpiexec -n 2"
 fi
+
+# test for hdf5 C++ exceptions
+if [[ "${CONDA_BUILD_CROSS_COMPILATION:-}" != "1" || "${CROSSCOMPILING_EMULATOR}" != "" ]]; then
+    $CXX ${CXXFLAGS} $RECIPE_DIR/testhdf5exc.cpp -o testhdf5exc -L$PREFIX/lib -lhdf5_cpp -lhdf5 -I$PREFIX/include
```
hmaarrfk, line 164:
```suggestion
    ${CXX} ${CXXFLAGS} ${LDFLAGS} $RECIPE_DIR/testhdf5exc.cpp -o testhdf5exc -lhdf5_cpp -lhdf5 -I$PREFIX/include
```

You can also follow the style and add the test to the test section:
https://github.com/xylar/hdf5-feedstock/blob/add-exception-catch-test/recipe/run_test.sh#L23

xylar: Agreed, I will move it if we decide we want to keep the test.

gillins:
Update: `valgrind` shows no errors running the `kealib` test on Linux x86_64 with `hdf5-1.14`, so nothing useful there....

xylar:
@gillins, no surprise.  The issue seems to be specific to clang.  My understanding based on some googling is that it's related to mixing up the system vs. the conda-forge (or in my google searches the homebrew) version of a core c++ library (can't recall which).  Anyway, seems both osx and clang specific.

gillins:
Ah ok, so might be a problem for other packages too? Unfortunately I no longer have a Mac to run other tests...

xylar:
I'm going to close this as I don't think it was helpful.

PR #254, opened 2025-03-18
Update to 1.14.6

sdebionne:
Closes #252 

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #255, opened 2025-03-21
1.14.6 + cmake

hmaarrfk:
Checklist


recipe/meta.yaml
```diff
@@ -1,6 +1,7 @@
-{% set version = "1.14.4" %}
-{% set maj_min_ver = ".".join(version.split(".")[:2]) %}
-{% set build = 5 %}
+{% set version = "1.14.6" %}
+{% set maj_min_patch_ver = "_".join(version.split(".")) %}
```
h-vetinari, line 2: I think it would be easier to just use `{{ version|replace(".", "_") }}` in the url, rather than introduce another variable.

hmaarrfk: sure.

hmaarrfk:
```suggestion
```

recipe/meta.yaml
```diff
@@ -37,23 +38,22 @@ package:
   version: {{ version }}
 
 source:
-  # Note: The additional "-3" at the end of the download link is an upstream accident and should go away with the next version.
-  url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-{{ maj_min_ver }}/hdf5-{{ version }}/src/hdf5-{{ version }}-3.tar.gz
-  sha256: 019ac451d9e1cf89c0482ba2a06f07a46166caf23f60fea5ef3c37724a318e03
+  url: https://support.hdfgroup.org/releases/hdf5/v{{ maj_min_ver }}/v{{ maj_min_patch_ver }}/downloads/hdf5-{{ version }}.tar.gz
```
hmaarrfk, line 41:
```suggestion
  url: https://support.hdfgroup.org/releases/hdf5/v{{ maj_min_ver }}/v{{ version|replace(".", "_") }}/downloads/hdf5-{{ version }}.tar.gz
```

recipe/meta.yaml
hmaarrfk: but you also need `1_14` and not just `1_14_6` this kind of churn isn't necessary.

Lets focus on the cmake imporvements and the version update.

recipe/meta.yaml
```diff
@@ -116,9 +113,6 @@ test:
         "h5pfc",
     ] %}
     {% endif %}
-    {% set hdf5_unix_cmds = h5_compilers + [
-        "h5redeploy",
-    ] %}
```
hmaarrfk, line 121: i believe that this is an automake style tool

hmaarrfk:
fingers crossed i skipped enough tests for linux.

Then it is off to osx cross compilation issues.

hmaarrfk:
@ajelenak is there some existing HDF5 build constants that you can point me to in order to bypass the checks for type sizes for OSX + arm64

hmaarrfk:
@ajelenak  @conda-forge/hdf5 i'm not sure if anybody else can jump in here. I mostly have trouble ensuring that the constants I define are getting passed through t the compilation settings.

I would love to release 1.14.6 in April, but I'm not sure I will have the investigate this.

LMK if anybody can help, otherwise the cmake transition may have to wait till 1.15 and we should merge #254 first.

hmaarrfk:
Hi @ajelenak thanks for chiming in.

> Where do you have the most problems? 

The bigges problems are getting through the cross compilation settings for OSX ARM. We build using a x86-64 runner, and target ARM. without emulation, alot of the "discovery code" that detects the word sizes just fails.

I don't understand the build system enough to create our own overrides.

> Modifying so many tests does not seem right. Is there a common underlying problem?

Generally speaking:
1. Underpowered runners
2. Parallel tests
3. Timeouts
4. Architecture Emulation
5. Buggy emulation of advanced instructions

make it so that high performance scientific code fails tests. Each skipped test is thematically related to one of the 5 points above and the tests were historically skipped on conda-forge. I've just had to port over the skips to cmake (I still have some skips to port).

ajelenak:
It looks like it may take a while to address these kinds of problems. If you have a quicker option to release 1.14.6, I suggest to use it.

recipe/build.sh
```diff
@@ -1,146 +1,76 @@
...
+  export H5_PAC_FC_ALL_REAL_KINDS="{4,8}"
+  export H5_H5CONFIG_F_NUM_IKIND="INTEGER, PARAMETER :: num_ikinds = 5"
+  export H5_H5CONFIG_F_NUM_RKIND="INTEGER, PARAMETER :: num_rkinds = 2"
+  export H5_H5CONFIG_F_IKIND="INTEGER, DIMENSION(1:num_ikinds) :: ikind = (/1,2,4,8,16/)"
+  export H5_H5CONFIG_F_RKIND="INTEGER, DIMENSION(1:num_rkinds) :: rkind = (/4,8/)"
+  export H5_PAC_FORTRAN_NATIVE_INTEGER_SIZEOF="                    4"
+  export H5_PAC_FORTRAN_NATIVE_INTEGER_KIND="           4"
+  export H5_PAC_FORTRAN_NATIVE_REAL_SIZEOF="                    4"
+  export H5_PAC_FORTRAN_NATIVE_REAL_KIND="           4"
+  export H5_PAC_FORTRAN_NATIVE_DOUBLE_SIZEOF="                    8"
```
hmaarrfk, line 40: @ajelenak are you saying that this kind of setting of environment variable does not ovewrite

these definitions?
https://github.com/HDFGroup/hdf5/blob/407caf1259dd03685b411f78321f823abd2666a6/config/cmake/HDF5UseFortran.cmake#L424

ajelenak: I don't know, have to ask our CMake expert. I have only set CMake variables on the command line (-D option).

Krande:
@hmaarrfk Intel just released IFX on conda-forge for windows in https://github.com/conda-forge/intel-compiler-repack-feedstock/pull/57. If you give me a few days I can add support for fortran in HDF5 for windows?

recipe/build.sh
hmaarrfk: That would be very helpful in terms of helping the community improve the state of HDF5 support

ajelenak: Setting env. variables is not going to affect those CMake variables. Options are setting them on the command line or in a custom cacheinit.cmake file which needs to be specified on the command line.

I know this is not the right place to ask but... is it possible to build conda packages somewhere else? I can build the library in a conda environment on my macOS arm64 laptop without any problem. If all the build tools are available to be installed in a conda environment, would it be possible to use GitHub actions to create conda packages?

hmaarrfk:
> @hmaarrfk Intel just released IFX on conda-forge for windows in [conda-forge/intel-compiler-repack-feedstock#57](https://github.com/conda-forge/intel-compiler-repack-feedstock/pull/57). If you give me a few days I can try to add support for fortran in HDF5 for windows?

ok sounds great.

I don't know if you want to use Cmake or autotools. Somebody needs to write a cmake patch for osx cross compilation.

Krande:
> > @hmaarrfk Intel just released IFX on conda-forge for windows in [conda-forge/intel-compiler-repack-feedstock#57](https://github.com/conda-forge/intel-compiler-repack-feedstock/pull/57). If you give me a few days I can try to add support for fortran in HDF5 for windows?
> 
> ok sounds great.
> 
> I don't know if you want to use Cmake or autotools. Somebody needs to write a cmake patch for osx cross compilation.

I have successfully added fortran support for windows using intel ifx for the `nompi` variant in the draft PR https://github.com/conda-forge/hdf5-feedstock/pull/256. I don't know enough about `impi` to know how to support that with ifx.

It should be safe to merge these changes into this PR.  

Regarding

> Somebody needs to write a cmake patch for osx cross compilation

I don't have access to any osx machine, so I might not be the best candidate for this. But I can probably find some time in the coming days to give it a shot (although I cannot make any promises).

hmaarrfk:
> I don't have access to any osx machine, so I might not be the best candidate for this. But I can probably find some time in the coming days to give it a shot (although I cannot make any promises).

do you have access to a linux machine? You could uninstall qemu and ensure that it can be compiled without emulation, that is the big challenge.

hmaarrfk:
> It should be safe to merge these changes into this PR.

if you could make it a single commit for me to cherry pick (keep the rerendering as a separate commit) that would make it easy to integrate here.

Krande:
> if you could make it a single commit for me to cherry pick (keep the rerendering as a separate commit) that would make it easy to integrate here.

I squashed the commits and made a single commit containing the changes. Hopefully that can be used: https://github.com/conda-forge/hdf5-feedstock/pull/256/commits/6e1eb998516b5626805ff4a570d8b50a0ad351ac


> do you have access to a linux machine? You could uninstall qemu and ensure that it can be compiled without emulation, that is the big challenge.

Yes, I have access to a linux machine. I can give it a shot

recipe/meta.yaml
```diff
@@ -72,6 +72,7 @@ requirements:
     - {{ stdlib('c') }}
     - {{ compiler('cxx') }}
     - {{ compiler('fortran') }}  # [not win]
+    - ifx_win-64                 # [win and mpi == 'nompi']
```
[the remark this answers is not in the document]
h-vetinari: Also, instead of hardcoding this in the recipe, it would be better to remove the selectors here and just add
```yaml
fortran_compiler:  # [win]
  - ifx            # [win]
```
in CBC

hmaarrfk:
I've manually checked:

* Gentoo -- https://packages.gentoo.org/packages/sci-libs/hdf5/dependencies
* Fedora -- https://src.fedoraproject.org/rpms/hdf5/tree/rawhide
* Debian -- https://salsa.debian.org/debian/hdf5/-/tree/master/debian/patches?ref_type=heads
* Arch -- https://gitlab.archlinux.org/archlinux/packaging/packages/hdf5/-/blob/main/PKGBUILD?ref_type=heads#L29 (lol at their notes)

The first three don't even depend on cmake, Arch references, but then uses autotools.

Ultimately, we would be trailblazing here, and unless we get help from upstream, I am tempted to move back to #254

hmaarrfk:
@mgorny is there any synergy with gentoo package manager in moving from autotools to cmake and ensuring cross compilation works without emulation? 

I think it needs some patches from some1 knowledgeable in cmake and I'm not sure if you are one that can help.

mgorny:
Hmm, I see they intend to drop autotools entirely, so I guess we'll have to do that. Yeah, I'll ping our folk and see if we can do something, either me or somebody else. That said, cross is not high on our priority list, so we'll probably only be addressing the problems that affect native builds.

mgorny:
Okay, I've asked around and we've tried the CMake build recently and it was really bad. Unfortunately, the problem is not CMake build per se but upstream deciding to make breaking changes to `h5cc` when building with CMake, which means reverse dependencies do not build anymore, e.g.:
- [gwyddion](https://bugs.gentoo.org/820158)
- [slurm](https://bugs.gentoo.org/813924)

And [upstream's response](https://github.com/HDFGroup/hdf5/issues/1814) is basically:

> CMake h5cc changes have been noted in the Release.txt at the time of change - archived copy should exist in the history files.

So basically they require you to update all the reverse dependencies before you switch to CMake. I don't know if any progress has been made on this, or if reverse dependency maintainers are even aware of the problem.

recipe/meta.yaml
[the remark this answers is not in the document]
Krande: @hmaarrfk I have no experience with impi and it failed to compile when I tested last time locally, so I just omitted it in my PR. But I will take a another look at adding impi support in a new PR based on the latest merged main and see if it I can figure it out.

mgorny:
Small update: I've looked around and it is possible that the original issue was fixed after all (or at least there are some commits that look like a fix) and the bug wasn't closed. I'm going to try to find some time to investigate this.

hmaarrfk:
@mgorny in terms of "urgency" i've decided to "punt this" until 1.15 because https://github.com/conda-forge/hdf5-feedstock/pull/254 was ready for 1 month and the author had been patiently waiting.

it would be great to be "prepared" for this, but it also seems like there are some efforts in making native OSX-arm compilation a possibility.

However, any help getting cmake working with cross compilation would definitely be helpful.

However, if we try to do it for the same HDF5 version, we need to make sure that the Autotools ABI is compatible with that of the cmake
https://conda-forge.org/docs/maintainer/knowledge_base/#moving-from-an-autotools-build-to-a-cmake-build



We also just got the gift of this beautiful bug: https://github.com/conda-forge/hdf5-feedstock/issues/258

so ...

mgorny:
Okay, some good news is that it seems that CMake is now installing usable h5cc. For 1.14.6, I have used the following pull requests:

1. https://github.com/HDFGroup/hdf5/pull/5361 — this fixes paths in h5cc
2. https://github.com/HDFGroup/hdf5/pull/5465 — here I fixed compatibility with POSIX sh (which regressed in the previous PR) and fixed quoting
3. https://github.com/HDFGroup/hdf5/pull/5467 — here I've added CMake vars to let people override the compiler used by `h5cc` and so on (the problem was originally noticed by openSUSE, https://build.opensuse.org/projects/science/packages/hdf5/files/hdf5.spec?expand=1)

This is my work-in-progress, and I'm in process of testing reverse dependencies now: https://github.com/gentoo/gentoo/pull/41704

Of course, I don't necessarily know what I'm doing.

mgorny:
Okay, so far the conclusion is that I haven't seen a single package failing with hdf5 CMake version that worked with the autotools version. I've found other issues, such as packages not using hdf5 when `--enable-hdf5` is passed due to messed up configure script, bugs in Gentoo packages, incompatibilities with new SWIG or GCC, but not a single failure resulting from use of CMake.

If you need to work on more specific CMake problems, please let me know what they are :-).

hmaarrfk:
> single package failing with hdf5 CMake version that worked with the autotools version. 

so are you saying that it is likely that CMake + autotools are ABI compatible in this case?
If you think so, we can likely do the cmake conversion before the 1.15.0 release. I was mostly worried about ABI compatibility

> CMake problems

For conda-forge, the biggest challenge stems from the (in)ability to override certain build time variables such as 
`PAC_FORTRAN_NATIVE_INTEGER_SIZEOF` https://github.com/HDFGroup/hdf5/blob/hdf5_1.14.6/config/cmake/HDF5UseFortran.cmake#L377

I am hoping to gather more information here, and make a plea upstream to have "more first class support for cross compilation", instead of asking downstream packagers to make their own patches.

It could be a win-win for all,

mgorny:
> so are you saying that it is likely that CMake + autotools are ABI compatible in this case?
> If you think so, we can likely do the cmake conversion before the 1.15.0 release. I was mostly worried about ABI compatibility

No clue about ABI compatibility. I just meant that `h5cc` regressions have been fixed, and they restored API / build system compatibility.

To be honest, I've taken ABI incompatibility at your word and just immediately marked it as incompatible.

mgorny:
There's at least incompatibility in Fortran libraries:

```diff
 *  SONAME:-libhdf5_f90cstub.so.310(64)
 *  SONAME:-libhdf5_hl_f90cstub.so.310(64)
 *  SONAME:-libhdf5_hl_fortran.so.310(64)
 *  SONAME:-libhdf5_tools.so.310(64)
 *  SONAME:+libhdf5hl_fortran.so.310(64)
```

`-` here are additional libraries installed by CMake, `+` by autotools.

hmaarrfk:
> To be honest, I've taken ABI incompatibility at your word and just immediately marked it as incompatible.

To be honest, the process of marking ABI compatible/incompatible after packages have been built is rather annoying at conda-forge.

So its easier to use the standard run_exports and rebuild.

The rebuilding process also helps "refresh" recipes so i'm not always opposed.

PR #299, opened 2026-03-24
fail when ctest fails, fix tests, backport vfd patch, stop ppc builds

minrk:
watching the logs in #295  I noticed there were actually a lot of test failures, but the ctest exit code was being ignored when the log file was shown.

This won't fix the errors (except for those due to timeouts with mpich, hopefully), but I expect it to fail.


hmaarrfk:
hmmm, so is hdf5-2 broken?

hmaarrfk:
Thanks the back port in #301 shows that this resolves the Linux 64 issues

hmaarrfk:
awesome. i would mark as ready and merge!

hmaarrfk:
Great work

PR #302, opened 2026-04-03
avoid exposing improper transitive dependency on zlib

minrk:
zlib shouldn't be required to link hdf5, but there's a `find_dependency` call in hdf5 cmake that I think shouldn't be there, leading to find_package(hdf5) failing if zlib dev files aren't present.

upstream issue: https://github.com/HDFGroup/hdf5/issues/6347

I haven't submitted a PR upstream because I'm not sure the logic here is correct for every hdf5 install (Windows, static, etc.), but I believe it is for ours (shared only)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set version = "2.1.0" %}
 {% set maj_min_patch_ver = "_".join(version.split(".")) %}
 {% set maj_min_ver = "_".join(version.split(".")[:2]) %}
-{% set build = 3 %}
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -97,8 +100,6 @@ test:
     - pkg-config                 # [not win]
     - cmake                      # [not win]
     - make                       # [not win]
-    # Need zlib to compile test programs
-    - zlib
```
hmaarrfk, line 101: Thank you

